PR #16, opened 2019-06-25
Typo fix

timClicks:
Will look to contribute more substantively when I've had a good look through the code!


indygreg:
Thank you for the fix, regardless how small!

PR #19, opened 2019-06-25
Fix tiny typo in readme

stigok:


indygreg:
Thank you for the fix!

PR #37, opened 2019-06-27
Fix bug when running setup.py

wwendler:
In the current build cmd.stdout.as_mut().unwrap(); on line 1136 fails and the output of running setup.py goes to the command line. I believe this change fixes things.


indygreg:
Yes, I think this will do the correct thing as well.

Issue #21, opened 2019-06-25
pip packaging rules should allow excludes


bsiegel:
I’m trying to package up a tool that can be installed from pip, so naturally I went with a `pip-install-simple` packaging rule. However I discovered that a transitive dependency (`include`) bundles both Python2 and Python3 modules side by side which was breaking the build. I needed to exclude the offending Python2 modules, but there is no exclude for the two pip-related packaging rules.

I solved this by switching to the `virtualenv` packaging rule which does allow specifying excludes, and just populating my virtualenv by running “pip install <packagename>”. However it would be nice if the exclude directive was allowed on the pip-related rules as well.

indygreg:
Thanks for the report! The features for various packaging rules could definitely be improved and made consistent. This is a legitimate feature request. I'll gladly accept a pull request if someone beats me to implementing this!

PR #40, opened 2019-06-27
pip packaging rules should allow excludes #21

Tipuch:
In reference to Issue #21 
- added option to exclude packages when packaging with pip install simple.


indygreg:
The technical implementation is correct and I will integrate this shortly. But it is lacking documentation, so I will follow it up with a change to the docs.

Thank you for the feature!

indygreg:
I landed this as `8f4442577` but Git/GitHub was not smart enough to recognize that. Closing.

Issue #28, opened 2019-06-26
Issue with Python distribution on Windows 10 with default init pyapp.


McSpidey:
Hi

Firstly, thanks for this great tool - it's exactly what I've wanted to join the worlds of Rust and Python and I can't wait to get it going!

pyoxidizer itself builds fine, both via crate and the github master build, but I'm having an issue with the stand-alone python aspect on Windows 10 it sems. Here's the logs trying to build the default in init pyapp example.

```
...pyapp>pyoxidizer build
no existing PyOxidizer artifacts found
processing config file ...pyapp\pyoxidizer.toml
resolving Python distribution...
downloading https://github.com/indygreg/python-build-standalone/releases/download/20190617/cpython-3.7.3-windows-amd64-20190618T0516.tar.zst
Python distribution available at ...pyapp\build\target\x86_64-pc-windows-msvc\debug\pyoxidizer\cpython-3.7.3-windows-amd64-20190618T0516.tar.zst
reading data from Python distribution...
thread 'main' panicked at 'unable to extract tar archive: Custom { kind: NotFound, error: TarError { desc: "failed to unpack `...pyapp\\build\\target\\x86_64-pc-windows-msvc\\debug\\pyoxidizer\\python.fd43554b5654\\python\\install\\Lib\\site-packages\\pip-19.1.1-py3.7.egg\\pip\\_vendor\\urllib3\\contrib\\_securetransport\\__pycache__\\low_level.cpython-37.pyc`", io: Custom { kind: NotFound, error: TarError { desc: "failed to unpack `python/install/Lib/site-packages/pip-19.1.1-py3.7.egg/pip/_vendor/urllib3/contrib/_securetransport/_` into `...pyapp\\build\\target\\x86_64-pc-windows-msvc\\debug\\pyoxidizer\\python.fd43554b5654\\python\\install\\Lib\\site-packages\\pip-19.1.1-py3.7.egg\\pip\\_vendor\\urllib3\\contrib\\_securetransport\\__pycache__\\low_level.cpython-37.pyc`", io: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." } } } } }', src\libcore\result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

```
---
Backtrace enabled Version:

```
pyoxidizer build
no existing PyOxidizer artifacts found
processing config file ...pyapp\pyoxidizer.toml
resolving Python distribution...
[...]

PR #41, opened 2019-06-27
Issue when extracting Python distributions on windows (long paths)

Tipuch:
This references issue #28
- added canonicalization so longer paths work on windows 10

### DIFF
--- a/pyoxidizer/src/pyrepackager/dist.rs
+++ b/pyoxidizer/src/pyrepackager/dist.rs
@@ -543,7 +543,8 @@ pub fn analyze_python_distribution_tar<R: Read>(
     let test_path = extract_dir.join("python").join("PYTHON.json");
     if !test_path.exists() {
         std::fs::create_dir_all(extract_dir).or_else(|_| Err("unable to create directory"))?;
-        tf.unpack(extract_dir)
+        let absolute_path = std::fs::canonicalize(extract_dir);
+        tf.unpack(&absolute_path.unwrap())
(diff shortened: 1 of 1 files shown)


indygreg:
This looks good to me and I will commit it shortly. But it may only fix the paths issue at extraction: there will likely be other operations that fail due to the long paths.

When I ran into the paths length issue locally several days ago, I attempted to work around by calling `canonicalize()` on the main `project_path` that was passed into `BuildContext::new()`. Essentially, this made every path prefixed with `\\?\`. Unfortunately, there was random functionality that didn't like the resulting `\\?\` prefix. This included things invoked by the `cc` crate either not accepting these paths or not escaping them properly. It seemed like a thorny problem. So I stopped using `canonicalize()`.

What I'm trying to say is that while this patch gets us one step closer to a solution, we may have to chase a long tail of problems caused by failing to call `canonicalize()`. I'm generally in favor of deferring `canonicalize()`, as no human uses the `\\?\` and I would strongly prefer for them to not appear in places like config files and log output.

Thank you for the fix!

Tipuch:
You're right, this seems like something that could pop up at different places in the code, from my limited understanding the standard library from rust handles long paths internally at several places in their code, so the pain points may be situated only when deferring to third party libraries that don't handle these kinds of issues (it should limit the number of pain points). They'll probably be found as things go, by the way is the part of the code concerned with using the cc crate vulnerable to long paths?

Issue #42, opened 2019-06-27
Support configuration of index-url for pip installs


goedelsoup:
While a global index-url configuration will work fine in most cases, it would be beneficial to supply this value on specific builds as well.

indygreg:
Good idea!

It's probably not worth the effort to duplicate every option that can be supplied to pip. So I think the way this should be implemented is as an `extra_args` option (or similar) that accepts an array of strings to pass to arguments to the invoked `pip` command. This is nearly infinitely extensible and should placate even the most discerning user's needs. It does allow people to shoot themselves in the foot by doing something unwanted. With great powers come great responsibility. This is a lesser evil than having to define dozens of options to control pip's various knobs.

indygreg:
Support for this just landed and will be in the next release. You can build PyOxidizer from the Git repository if you want the feature immediately.

PR #52, opened 2019-06-28
Support extra_args for pip install simple

Tipuch:
This pull request references #42 
- added extra_args to pip install simple
- added documentation

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -693,6 +693,13 @@ rule.
 
    Default is ``true``.
 
+``excludes`` (array of string)
+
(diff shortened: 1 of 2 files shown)


Tipuch:
I've looked at the TODO for piping stderr to stdout, it seems to be hard to do with the standard library, a crate named duct seems to be adapted to what we're looking for, I wasn't sure if you wanted another dependency or if there was a simpler way to do it so I kept it out of the pull request for now.

Tipuch:
I just noticed the extra commits, I'm a bit new to this, I'll clean this up.

PR #54, opened 2019-06-28
Support extra_args for pip install simple

Tipuch:
This pull request references #42

 -  added extra_args to pip install simple
 -  added documentation

I've looked at the TODO for piping stderr to stdout, it seems to be hard to do with the standard library, a crate named duct seems to be adapted to what we're looking for, I wasn't sure if you wanted another dependency or if there was a simpler way to do it so I kept it out of the pull request for now.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -705,6 +705,10 @@ rule.
    Where to package resources associated with this rule.
    See :ref:`install_locations`.
 
+``extra_args`` (optional array of string)
+
+   An array of arguments added to the pip install command.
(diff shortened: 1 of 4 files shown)


indygreg:
Great patches! Thanks for implementing this!

Regarding the extra crate dependency, that's fine: `pyoxidizer` can depend on pretty much anything. I mainly care about the dependencies leaking into `pyembed`, as those get shipped with built binaries and I want to keep that set as small as possible.

PR #57, opened 2019-06-29
TODOS: stderr to stdout

Tipuch:
- added duct as a dependency to allow more control on commands
- added stderr_to_stdout


indygreg:
I want to acknowledge the existence of this PR. I'm not trying to ignore it, I just wanted to focus on feature and bug work this weekend.

This patch is likely severely bit rotted as a result of the work I did this weekend. I'm sincerely sorry about that and I'm willing to rebase it myself. I'll try to do that in the next ~24 hours unless you beat me to it.

Tipuch:
Holy rebase... I'm not sure this is entirely right.

Tipuch:
by the way, no worries about other priorities I totally get that. I just hope I didn't mess up the commit history too bad, I think the changes are the correct one (it may need another pair of eyes to confirm it though)

pyoxidizer/src/pyrepackager/packaging_rule.rs
```diff
@@ -839,24 +839,18 @@ fn resolve_pip_install_simple(
...
     }
 
-    // TODO send stderr to stdout.
-    let mut cmd = std::process::Command::new(&dist.python_exe)
-        .args(&pip_args)
-        .envs(&extra_envs)
-        .stdout(std::process::Stdio::piped())
-        .spawn()
+    let handler = cmd(&dist.python_exe, &pip_args)
+        .full_env(&extra_envs)
```
indygreg: `.full_env()` will replace the environment variables with the set specified. This is different from what `.envs()` did before, which adds multiple environment variables to the existing set. We need to preserve the old behavior.

pyoxidizer/src/pyrepackager/packaging_rule.rs
```diff
@@ -839,24 +839,18 @@ fn resolve_pip_install_simple(
...
+        .stdout_capture()
+        .start()
         .expect("error running pip");
-    {
-        let stdout = cmd.stdout.as_mut().unwrap();
-        let reader = BufReader::new(stdout);
+    
+    let output = handler.wait().unwrap();
+    let stdout = &output.stdout;
+    warn!(logger, "{}", std::str::from_utf8(&stdout).unwrap());
```
indygreg: Can `duct` stream output? I don't like buffering all the output until the process finishes because it can suppress output from long-running commands until much later after it occurs. This gives the feeling of slow execution, makes it harder to understand timings, etc.

pyoxidizer/src/pyrepackager/packaging_rule.rs
Tipuch: they don't seem to have implemented this (according to the docs) I think copying the parent's environment variables and chaining them with the extra_envs variables should be satisfactory.

pyoxidizer/src/pyrepackager/packaging_rule.rs
Tipuch: They have an example of it in the docs, I'll make that work (probably tomorrow)

indygreg:
I wanted to let you know that I haven't forgotten about this PR. I've just had other things occupying my life recently...

Tipuch:
Don't worry about it, I've had a pretty busy week as well, I'll probably fix the remaining conflicts today or tomorrow.

Tipuch:
This PR looks like it is dead since things have changed significantly, I'll just close it.

Tipuch:
I have to admit I'd rather have known earlier on that this wasn't going to be implemented. It feels like wasted time.

PR #58, opened 2019-06-29
Fix grammar in first FAQ sentence

nealmcb:
Upon further reading, I guess if you really want to make the phrase "turn Python into distributable" into an adjective, the other alternative is to put dashes in there. But as is, it was so distracting and clear to me that this was a simple typo that I made a whole fork 8)


indygreg:
Yes, your phrasing is much better! (I wrote this paragraph a few minutes after waking up and didn't have enough tea in me. Look at the commit history for these lines to see how bad the initial version was!)

PR #59, opened 2019-06-29
Clarify context: building portable binaries

nealmcb:
At first it seemed like the "execution environment for binaries" would defeat the whole purpose of distributing a portable binary.
Trying to clarify that.


indygreg:
Thanks for the wordsmithing!

PR #62, opened 2019-07-01
repair: pip-requirements-file type args must be requirements_path not…

zhuima:
… path

   [[packaging_rule]]
   type = "pip-requirements-file"
   path = "/home/gps/src/myapp/requirements.txt"

to

   [[packaging_rule]]
   type = "pip-requirements-file"
   requirements_path = "/home/gps/src/myapp/requirements.txt"


indygreg:
Good catch! Thanks for the fix.

PR #67, opened 2019-07-01
Avoid panic from unwrap() when no arg for analyze

mwm126:


indygreg:
Nice! Thanks for the fix!

PR #86, opened 2019-07-07
Get rid of benign warnings about trait objects and dyn

svenstaro:
Before:
```
warning: trait objects without an explicit `dyn` are deprecated
```
Now: fixed :D


indygreg:
Thanks for the improvement!

PR #174, opened 2019-11-06
Add stdlib excludes

jayvdb:
Allows exclusion of stdlib modules such as distutils, pip, ensurepip
which are large and often unnecessary in a standalone binary.


indygreg:
This landed as 35ab90d.

Issue #26, opened 2019-06-26
pyca/cryptography: cffi build and __file__


alexzorin:
Hello,

Very cool project - it's what I wish Oracle's Graal could do for Python.

I'm wondering whether you could give a hint for using [pyca/cryptography](https://pypi.org/project/cryptography/). I'm actually trying to see how hard it would be to build Certbot, but this is the minimal repro:

```toml
[[packaging_rule]]
type = "pip-install-simple"
package = "cryptography==2.7"

[[embedded_python_run]]
mode = "eval"
code = "from cryptography.hazmat.backends import default_backend; default_backend()"
```

Runtime error:

    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "cryptography.hazmat.backends", line 15, in default_backend
      File "cryptography.hazmat.backends.openssl", line 7, in <module>
      File "cryptography.hazmat.backends.openssl.backend", line 18, in <module>
      File "cryptography.x509", line 8, in <module>
      File "cryptography.x509.base", line 16, in <module>
      File "cryptography.x509.extensions", line 19, in <module>
      File "cryptography.hazmat.primitives.constant_time", line 11, in <module>
    ModuleNotFoundError: No module named 'cryptography.hazmat.bindings._constant_time'

Looking [at how pyca/cryptography seems to build this module](https://github.com/pyca/cryptography/blob/3bab819c558ead888825e33f8df3f17707ff2c20/src/_cffi_src/build_constant_time.py#L13), I believe this is a variation of the `__file__` problem described in the docs.

I've not had much success with applying the advice from the docs. I think this might be because of the use of `__file__` as part of the cffi build process rather than something happening at runtime. I'm not that familiar with the Python runtime in general :man_shrugging:.

If this is too complex or impossible at the moment, that's okay too! Thanks.

seanjensengrey:
I'd like to take a look at the specifics of how the file system encapsulation is done, but if we interpose open/read/etc we could reroute some calls for resolving files to an inproc function and read only segment in the executable. Stuff like late loading of resources (even non-python resources) would then _just work_. Other calls that are outside of a certain root would be passed to libc/syscall and continue to function as expected.

indygreg:
I'm pretty certain the `cryptography` package uses C extensions. As things are currently implemented, C extensions *may* be built during PyOxidizer packaging. But there is no guarantee that will work, those binaries will be portable, or that the extension files will be recognized as such and installed in the proper location.

[...]

PR #181, opened 2019-11-08
distutils: Use ext_package in built ext names

jayvdb:
distutils setup() accepts an ext_package value for the name
of the package for a group of built extensions.

Fixes https://github.com/indygreg/PyOxidizer/issues/26


indygreg:
This landed as c5efc0a.

PR #185, opened 2019-11-09
Add extra_env to pip packaging rules

jayvdb:
The `extra_env` argument for the setup.py rule is also
applicable for the two pip rules.


indygreg:
This landed as 08bd4b6.

PR #183, opened 2019-11-08
Rename PyInit syms to avoid clashes

jayvdb:
Built extensions in packages often have common names
like speedups, utils, _objects, cpython, etc. which
reside inside the package namespace.
The compiled extensions each have a PyInit_<module>
which needs to be renamed to PyInit_<pkg>_<module>
to avoid clashes when combined into a static binary.

Fixes https://github.com/indygreg/PyOxidizer/issues/169

PR #192, opened 2019-11-15
WIP: Rename stdlib _queue init if gevent present

jayvdb:
This depends on https://github.com/indygreg/PyOxidizer/pull/183


indygreg:
@jayvdb could you please provide more context on why this is needed? I just want to understand why we need to do this...

jayvdb:
@indygreg this PR is irrelevant until https://github.com/indygreg/PyOxidizer/pull/183 is merged.  And the problem this PR fixes is documented fairly well on that PR, but it is only a workaround until rust crate `object` is enhanced, but I also have been working on other workarounds instead of this, so I will mark this PR as a WIP as I think my new workarounds (not yet pushed) are better.

jayvdb:
This special case has been fixed upstream in https://github.com/gimli-rs/object/issues/159 , so doesnt need a workaround here.  Just need to wait for the next release of `object`.

PR #200, opened 2019-11-19
CI: Enable PipInstallSimple rule

jayvdb:
Closes https://github.com/indygreg/PyOxidizer/issues/206

### DIFF
--- a/ci/azure-pipelines-template.yml
+++ b/ci/azure-pipelines-template.yml
@@ -49,12 +49,12 @@ jobs:
 
   - ${{ if ne(parameters.name, 'Windows') }}:
     - script: |
-        cargo run --bin pyoxidizer -- init --python-code 'print("hello, world")' ~/pyapp
-        cargo run --bin pyoxidizer -- run ~/pyapp
+        cargo run --bin pyoxidizer -- init --pip-install appdirs ~/pyapp
+        cat ci/pyapp.py | cargo run --bin pyoxidizer -- run ~/pyapp
       displayName: Build Oxidized Application
 
   - ${{ if eq(parameters.name, 'Windows') }}:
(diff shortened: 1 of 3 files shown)


ci/azure-pipelines-template.yml
```diff
@@ -49,12 +49,12 @@ jobs:
...
+        cargo run --bin pyoxidizer -- init --pip-install appdirs ~/pyapp
+        cat ci/pyapp.py | cargo run --bin pyoxidizer -- run ~/pyapp
       displayName: Build Oxidized Application
 
   - ${{ if eq(parameters.name, 'Windows') }}:
     - script: |
-        cargo run --bin pyoxidizer -- init --python-code 'print("hello, world")' %HOME%/pyapp
-        cargo run --bin pyoxidizer -- run %HOME%/pyapp
+        cargo run --bin pyoxidizer -- init --pip-install appdirs %USERPROFILE%/pyapp
+        cat ci/pyapp.py | cargo run --bin pyoxidizer -- run %USERPROFILE%/pyapp
```
jayvdb: Need to also fix https://github.com/indygreg/PyOxidizer/issues/208 before this is a good idea, otherwise CI errors will be not seen.

indygreg:
I manually applied these commits. I dropped the changes to refactor the CI to use a standalone file because I don't think it adds sufficient value to include.

Regarding CI, I would prefer for as much functionality as possible to be tested in `cargo test`. And if `cargo test` is not sufficient for command-line level tests, then we should establish a dedicated script to run that invokes `pyoxidizer` and test with that: I do not want low-level test code defined in YAML files because the CI configuration should not be the source of truth on how tests are defined and how to invoke them.

PR #205, opened 2019-11-20
CI: Run simple built extension

jayvdb:


jayvdb:
Initially this was installing `simplejson`, and importing it on non-Windows only - green.

indygreg:
OK. The reasons for the CI change show more promise in this PR. I think I will change the CI to use the Python script because more value is demonstrated.

indygreg:
I manually applied this. As part of doing so, I pinned the versions of the installed packages so behavior is more deterministic.

Issue #141, opened 2019-10-09
test subcommand


jayvdb:
It would be nice if pyozidizer had a test command that was the equivalent of `setup.py test` (I know setuptools is trying to kill it, and that it will only work for a subset of packages, but ~50% of the core packages provided by a Unix distro will support it.  And adding an equivalent of `setup.py pytest` would provide much higher compatibility)

A very useful part of Nuitka is that it can build Python packages as loadable shared libraries.  This means that each dependent package can then be tested using the normal test suite of the package, using whatever test runner is needed, such as nose or pytest, and the test runner doesnt need to have been compiled with Nuitka.  This doesnt guarantee that a standalone built with Nuitka would also pass the test suite of those packages, but it is a helpful smoke test to catch dependencies doing something strange in a new release which would break the standalone in a hard to QA / detect way.

I can achieve an even better result with PyOxidizer with some fiddling. e.g. the following relevant part of `pyoxidizer.toml`

```toml
[[packaging_rule]]
type = "pip-install-simple"
package = "pytz"

# Add tests which are not installed
[[packaging_rule]]
type = "package-root"
path = "/path/to/sourcetree/pytz-2019.3"
packages = ["pytz.tests"]
```

```py
$ /path/to/pytz
>>> import unittest
>>> unittest.main('pytz.tests.test_lazy')
.......................................
----------------------------------------------------------------------
Ran 39 tests in 0.010s

OK

$ /path/to/pytz
>>> import unittest
>>> unittest.main('pytz.tests.test_tzinfo')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "unittest.main", line 70, in __init__
  File "pytz.tests.test_tzinfo", line 40, in <module>
  File "pytz", line 172, in timezone
  File "pytz", line 198, in _case_insensitive_zone_lookup
  File "pytz.lazy", line 101, in _lazy
  File "pytz", line 1104, in <genexpr>
  File "pytz", line 114, in resource_exists
  File "pytz", line 95, in open_resource
NameError: name '__file__' is not defined
```

boom.  There is the failure in https://github.com/indygreg/PyOxidizer/issues/91 .

Roughly the same toml is needed for `appdirs`, excepting its tests are in a `test` package which isnt installed.

If that could be wrapped up into a test command, it would be easier to determine which packages are likely broken when built with PyOxidizer, and also provide a way to quickly verify that those bugs have been fixed.

[...]

Issue #140, opened 2019-10-09
importlib_metadata.PackageNotFoundError: importlib_metadata


jayvdb:
When building a package including importlib_metadata, it fails.

During the build
```console
...
package importlib_metadata-0.23-py3.7.egg-info does not exist; excluding resources: ["PKG-INFO", "SOURCES.txt", "dependency_links.txt", "installed-files.txt", "requires.txt", "top_level.txt"]
package more_itertools-7.2.0-py3.7.egg-info does not exist; excluding resources: ["PKG-INFO", "SOURCES.txt", "dependency_links.txt", "installed-files.txt", "top_level.txt"]
package zipp-0.6.0-py3.7.egg-info does not exist; excluding resources: ["PKG-INFO", "SOURCES.txt", "dependency_links.txt", "installed-files.txt", "requires.txt", "top_level.txt"]
```
resolved 4 embedded resource files across 2 packages: {
    "importlib_metadata.docs": {
        "changelog.rst",
        "index.rst",
        "using.rst",
    },
    "importlib_metadata.tests.data": {
        "example-21.12-py3-none-any.whl",
    },
}
```

```py
>>> import importlib_metadata
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "importlib_metadata", line 547, in <module>
  File "importlib_metadata", line 509, in version
  File "importlib_metadata", line 482, in distribution
  File "importlib_metadata", line 187, in from_name
importlib_metadata.PackageNotFoundError: importlib_metadata
```

Could this be because its setup.py is
```
from setuptools import setup

setup(use_scm_version=True)
```

And setup.cfg includes `setup_requires=setuptools-scm`.

I tried adding setuptools-scm to the project first, and then it is added to the list
```
package importlib_metadata-0.23-py3.7.egg-info does not exist; excluding resources: ["PKG-INFO", "SOURCES.txt", "dependency_links.txt", "installed-files.txt", "requires.txt", "top_level.txt"]
package more_itertools-7.2.0-py3.7.egg-info does not exist; excluding resources: ["PKG-INFO", "SOURCES.txt", "dependency_links.txt", "installed-files.txt", "top_level.txt"]
package setuptools_scm-3.3.3-py3.7.egg-info does not exist; excluding resources: ["PKG-INFO", "SOURCES.txt", "dependency_links.txt", "entry_points.txt", "installed-files.txt", "top_level.txt", "zip-safe"]
[...]

PR #211, opened 2019-12-02
pyembed: Add private finder._package_list

jayvdb:
In order to programatically inspect and validate the contents
of the executable from Python code, the internal package list
needs to be exposed.

This also provides the ability for Python code to implement
package listing support in the various frameworks such as
pkg_resources and importlib.metadata's find_distributions.

Related to https://github.com/indygreg/PyOxidizer/issues/140
Related to https://github.com/indygreg/PyOxidizer/issues/141

### DIFF
--- a/pyoxidizer/src/pyembed/importer.rs
+++ b/pyoxidizer/src/pyembed/importer.rs
@@ -453,6 +453,23 @@ py_class!(class PyOxidizerFinder |py| {
             Ok(py.None())
         }
     }
+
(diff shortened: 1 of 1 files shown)


jayvdb:
As effective would be to create a compliment of `sys.builtin_module_names` for non-built packages, so the two combined contain all of the names needed to know what was included in the binary.

indygreg:
The `oxidized_importer` module now has a robust API for querying state. This PR is no longer needed.

PR #213, opened 2019-12-08
starlark: implement PythonDistribution.pip_requirements_file()

houqp:
This reimplements deleted PipRequirementsFile.

### DIFF
--- a/pyoxidizer/src/starlark/python_distribution.rs
+++ b/pyoxidizer/src/starlark/python_distribution.rs
@@ -251,6 +251,43 @@ starlark_module! { python_distribution_module =>
     }
 
     #[allow(clippy::ptr_arg)]
+    PythonDistribution.pip_requirements_file(env env, this, path, extra_envs=None) {
+        required_str_arg("path", &path)?;
+        optional_dict_arg("extra_envs", "string", "string", &extra_envs)?;
+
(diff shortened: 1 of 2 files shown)


indygreg:
Thank you for the contribution!

However, I think `PythonDistribution.pip_install()` is flexible enough to work with requirements files because it accepts an arbitrary list of arguments to pass to `pip install`. What do you think?

houqp:
Ha, I see, this makes sense. Will switch to use pip_install :)

indygreg:
Also, you may want to stay away from the `main` branch right now: things are very broken due to the partial transition to Starlark. I made a lot of progress today, however. But it is still not possible to do advanced packaging on the `main` branch.

houqp:
I figured that's the case by reading the code :)

Do you have a rough estimation on when master will be ready for testing? I ran into some issues with stable release, so was trying to see if I can reproduce with latest code.

BTW, I am excited for Starlark support, really looking forward to it!

indygreg:
It's probably another solid day of hacking to get Starlark back to feature parity with where we were before. When I'll find that day, I'm not sure. The new code actually has reasonable test coverage. So I'm optimistic that things will be in a much better state once the transition is complete. Then I can focus on more features and bug fixes.

houqp:
Got it. Let me know if there is anything I could help coding wise to speed up the starlark migration.

PR #225, opened 2020-02-13
Fix read_package_root example in .bzl template

pharaujo:
The example configuration was invalid as it was missing a `)`


indygreg:
Thanks for the bug fix - I'll merge this in shortly.

indygreg:
I rebased and landed as 1d811896b1c76e48b849feaf7d16ba943a6ea2a4 but GitHub isn't smart enough to detect that. So closing the PR manually.

PR #247, opened 2020-05-04
pyoxidizer: upgrade to reqwest 0.10 from 0.9

durin42:
We have to use the `blocking` feature to get a drop-in blocking-API
client, but then it was a clean upgrade. This removes several
duplicates in the dependency graph shown by `cargo tree --duplicates`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,6 +83,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bit-set"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


durin42:
Looks like this is failing in the guts of the `bytes` crate somehow, which I find mysterious.

indygreg:
Yeah, it is failing on the Rust 1.36 build. I have wanted to keep the project building with a Rust 2018 Edition era release for the Linux distros that move at a glacial pace. I had already conceded the bump from 1.32 to 1.36 to placate starlark-rust. I'm tempted to bump further because I haven't been able to upgrade various packages (including `reqwest`) to latest versions for a few months now due to 1.36 requirement. It is quite annoying and it appears the only viable strategy in the Rust ecosystem is to live on latest stable or not take meaningful updates.

indygreg:
I just bumped the minimum Rust version to 1.39 and attempted to retrigger CI here. Although I'm unsure if it actually triggered a run against the latest commit on `main`. Would you mind rebasing and pushing again to give it a kick?

durin42:
Pushed again. I'm certainly open to not upgrading rust if we'd rather stay on the old version, but it does feel like that's an option that's slowly slipping away.

PR #268, opened 2020-06-12
Add recommendation to use pip-licenses

johnthagen:
[`pip-licenses`](https://github.com/raimon49/pip-licenses) can help audit Python package licenses, by generating a license report for a set of dependencies.


johnthagen:
I presume the failing CI is unrelated to this change?

johnthagen:
Hmm, I now see that `pyoxidizer python-distribution-licenses` exists, perhaps that should be recommended instead?

indygreg:
I rebased and pushed as c7cc6bd. Thank you for the contribution!

PR #270, opened 2020-06-17
Fix typo on rust_rust_code page.

stuhood:


indygreg:
I rebased and landed as 43e98e0. Thanks for the fix!

PR #271, opened 2020-06-18
Re-run the pyembed build script if the config file contents have changed

stuhood:
When editing the `pyoxidizer.bzl` config file, `pyembed`'s `build.rs` will not re-run in `build-mode-pyoxidizer-exe` mode.

Further changes in this area are likely necessary though: when the config file is executed, it can walk various portions of the real filesystem to discover what it needs to re-run for (via e.g. `read_package_root` and others).


indygreg:
I rebased and landed as 24c3677.

I suspect this regressed from an earlier release, as I'm pretty sure this used to work!

PR #272, opened 2020-06-18
Render the extension name when we fail to embed something.

stuhood:
When a `PythonExtensionModule` is discovered via a batch-discovery method like `read_package_root` or `pip_install`, "which" module is causing the trouble is not obvious. Rendering the extension module name gives information about the culprit (although it will not pinpoint a file on disk or `requirements.txt` entry precisely).


indygreg:
I rebased and pushed as 7c3b3f6.

PR #285, opened 2020-09-15
PyO3 supports stable Rust as of version 0.11.0 (Jun 28, 2020)

trevyn:


indygreg (approved):
I didn't realize PyO3 worked with stable Rust now. Niiiiiice!

PR #305, opened 2020-10-28
cleanup: replace copy_dir crate with fs_extra crate

durin42:
The former appears to be largely abandoned and depends on an ancient
walkdir crate. The latter appears to be actively maintained, and this
prunes the ancient walkdir from our dependencies.


indygreg (approved):
Yay. Thanks.

PR #308, opened 2020-11-03
Passing dist to make_exe in eample

### LINKED ISSUE #306: PyPI example not working - pyoxidizing pyflakes
When trying build a pyflakes package like explained in [Packaging an Application from a PyPI Package](https://pyoxidizer.readthedocs.io/en/stable/packaging_python_files.html#packaging-an-application-from-a-pypi-package) I'm getting the following error message:

```
$ pyoxidizer build
resolving 1 targets
resolving target install
resolving target exe
resolving target dist
error[CF05]: Extraneous parameter passed to function call
  --> ./pyoxidizer.bzl:52:1
   |
52 | resolve_targets()
   | ^^^^^^^^^^^^^^^^^ Extraneous parameter in function call


error: Extraneous parameter passed to function call
```

The same applies to `pyoxidizer run`. If I understood correctly, replacing the `make_exe` function with the example function should be enough. The same happens on an Ubuntu installation of mine. I'm a bit clueless. Is this a bug or have I done something wrong? Thanks!

Here are the versions:
```
PyOxidizer 0.9.0
Python 3.8.5
openSUSE Tumbleweed
```

brejoc:
Seems like dist needs to be passed to `make_exe()`, which is a bit weird, because in the line below `dist` gets overwritten. But if `dist` is not passed, you get the error from #306. 

Fixes #306

### DIFF
--- a/docs/packaging_python_files.rst
+++ b/docs/packaging_python_files.rst
@@ -131,7 +131,7 @@ comments removed for brevity):
 
 .. code-block:: python
 
-   def make_exe():
+   def make_exe(dist):
        dist = default_python_distribution()
 
        policy = dist.make_python_packaging_policy()
--- a/docs/packaging_python_files.rst
+++ b/docs/packaging_python_files.rst
@@ -183,7 +183,7 @@ following:
 
 .. code-block:: python
 
-   def make_exe():
+   def make_exe(dist):
        dist = default_python_distribution()
 
        policy = dist.make_python_packaging_policy()


indygreg (approved):
Thank you for fixing a bug in the documentation!

Your PR summary is correct that it is weird that `dist` is overwritten despite being passed in. That's because it should not be overwritten!

I'll apply your PR locally, make the trivial change to drop the bad lines and push it.

indygreg:
I pushed this as fea05621 but GitHub didn't figure that out. Thank you again for your contribution!

PR #313, opened 2020-11-05
environment: remove runtime dependency on git2 crate

durin42:
For reasons that are complicated, I'm blocked on importing the git2
crate to our monorepo at work, but I still want to be able to start
working on bazel rules for pyoxidizer binaries. This lets users avoid
the git2 dependency in the pyoxidizer tool so they can cut a
significant part of the dependency tree.

### DIFF
--- a/pyoxidizer/Cargo.toml
+++ b/pyoxidizer/Cargo.toml
@@ -32,7 +32,6 @@ codemap-diagnostic = "0.1"
 duct = "0.13"
 fs2 = "0.4"
 fs_extra = "1.2"
-git2 = "0.13"
 glob = "0.3"
 goblin = "0.2"
 handlebars = "3.0"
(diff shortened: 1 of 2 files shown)


indygreg:
As discussed via video call, I think I can do better than this patch and remove the run-time `libgit2` dependency completely. So there's a good chance I close this today if I'm successful in the alternative, more involved change.

indygreg:
I rewrote all this logic in 84f44126..562ac88d. build.rs now contains most of the logic for resolving the location of the PyOxidizer install.

If you want to make libgit2 optional for build.rs as well, that patch should be pretty easy. And the new environment variables I added to the build.rs may make it easier for you to point PyOxidizer's autogenerated Rust projects at a different location. I know you didn't ask for this feature. But I wouldn't be surprised if you or some other user needs it. Hermetic builds and all of that.

PR #323, opened 2020-11-12
cleanup: rename new-pyoxidizer.bzl to new-pyoxidizer.bzl.hbs

durin42:
I keep getting blocked by buildifier[0] presubmit checks because the template
isn't a valid bzl file. It appears to be pretty usual to use .hbs as a
suffix on handlebars files, so rename the .bzl file to .bzl.hbs to
avoid the presubmit trying to force us to format a templated file.

Sigh.

0: https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md


indygreg:
Sigh.

Thanks.

PR #343, opened 2020-12-19
Fix #337: Add __path__ filtering to OxidizedFinder's pkgutil.iter_modules support

### LINKED ISSUE #337: Complete OxidizedFinder.iter_modules() support with __path__ filtering
I'm working on a PR or two that I'll submit hopefully this week. Wanted to get this ticket filed now to put it on the radar and provide a place for discussion before I get around to pushing code.

`OxidizedFinder.iter_modules()` was first added in #253. [According to the docs](https://pyoxidizer.readthedocs.io/en/latest/oxidized_importer_behavior_and_compliance.html#pkgutil-compatibility)

> `OxidizedFinder` does not populate `sys.path_importer_cache`, so path-based filtering via `pkgutil.iter_modules(path=...)` will not work like it does with the standard library’s importer.

I propose to fix this by adding an optional `path` argument to the beginning of `OxidizedFinder`'s constructor's argument list. Then `OxidizedFinder` itself can be appended to [`sys.path_hooks`]. Like [`FileFinder`]'s `path` argument, `OxidizedFinder`'s `path` argument would restrict the `OxidizedFinder` instance's searches to a given path. Unlike [`FileFinder`], a path beginning with [`sys.executable`] `+` [`os.sep`] would restrict the `OxidizedFinder` instance to searching for indexed resources embedded in the executable in the package whose [`__path__`] was set to that value.

I'm still pretty new to Rust and learning my way around the `pyembed` crate, so I'm very open to suggestions about how to do this or alternative approaches.

For context, this would support [`pkgutil.iter_modules`](https://docs.python.org/3/library/pkgutil.html#pkgutil.iter_modules). My main goal is to work around https://code.djangoproject.com/ticket/30950

[`sys.path_hooks`]: https://docs.python.org/3/library/sys.html#sys.path_hooks
[`__path__`]: https://pyoxidizer.readthedocs.io/en/latest/oxidized_importer_behavior_and_compliance.html#path-module-attribute
[`os.sep`]: https://docs.python.org/3/library/os.html#os.sep
[`sys.executable`]: https://docs.python.org/3/library/sys.html#sys.executable
[`FileFinder`]: https://docs.python.org/3/library/importlib.html#importlib.machinery.FileFinder

wkschwartz:
This PR is a rewrite of and supersedes #340, and it fixes #337.

# Summary

To make `pkgutil.iter_modules(pkg.__path__)` list only modules in `pkg`, I've added the `OxidizedFinder.path_hook` method, and, based on existing settings in `OxidizedPythonInterpreterConfig`, the `OxidizedFinder` instance in `sys.meta_path` has its bound `path_hook` method added to `sys.path_hooks`. When `pkgutil.iter_modules` passes `pkg.__path__` to `path_hook`, `path_hook` returns an `importlib.abc.PathEntryFinder` object whose `iter_modules` method searches the `OxidizedFinder` instance for embedded resources whose names identify them as being in `pkg`.

# Review

I think this is ready to go as long as you agree with me that none of the [known bugs](#known-bugs) are blockers. All tests pass on my Mac except tests that were failing before in `main` at 2b4e66406f4fafdd7e09bc469878b239dcd55880. I'll run tests on Windows this weekend or early next week.

# Known bugs

Multiple `OxidizedFinder` instances' `path_hook`s being in `sys.path_hooks` would interfere with each other.  The first one in `sys.path_hooks` wins control over all paths under `sys.executable`.

The `PathEntryFinder` that `OxidizedFinder.path_hook` returns does not (yet) implement legacy methods `find_loader` and `find_module`, which were deprecated in Python 3.4. They wouldn't be hard to write, but I'm having trouble convincing myself there's any point. I have confirmed that nothing in the standard library relies on those methods. If we really want those methods, I think they can be added in future PRs, and we can crib [Python's implementations](https://github.com/python/cpython/blob/550e4673be538d98b6ddf5550b3922539cf5c4b2/Lib/importlib/abc.py#L95-L127).

`OxidizedFinder.path_hook` panics when a path contains an unpaired surrogate after the `sys.executable` part of the path because of dgrunwald/rust-cpython#246. I added a `unittest.expectedFailure` test case for a `UnicodeDecodeError` being raised instead.

My algorithm for separating the `sys.executable` part of the path from the `pkg/mod` part of the path has some trash strewn about it. The worst bit is where I make a couple of allocations (`.collect<PathBuf>`) where really none are needed. Open to suggestions about how to tighten that up.

# UPDATE (12/22/20): Backwards incompatible change to achieve compatibility with CPython standard library

After some further testing I noticed that `pkgutil.iter_modules()` (i.e., with no `path` argument) was behaving differently when an `OxidizedFinder` was present on `sys.meta_path`. The standard library (vanilla CPython) yields only top-level modules (e.g., `importlib` but not `importlib.abc`). Moreover, the standard library only yields the last component of modules' `__name__`s:

```pycon
>>> import importlib, pkgutil
>>> [name for finder, name, ispkg in pkgutil.iter_modules(importlib.__path__)]
['_bootstrap', '_bootstrap_external', 'abc', 'machinery', 'metadata', 'resources', 'util']
```

02b9c2504960b3e53ea52eb2cf4224958f818d92 modifies `OxidizedFinder.iter_modules` and `OxidizedFinder().path_hook(path).iter_modules` to conform to the standard library's behavior. The documentation says almost nothing about a finder's `iter_modules` method, so the standard library is our best source of expected behavior.

### DIFF
--- a/pyembed/src/importer.rs
+++ b/pyembed/src/importer.rs
@@ -909,7 +909,9 @@ impl OxidizedFinder {
             None
         };
 
-        resources_state.pkgutil_modules_infos(py, prefix, state.optimize_level, |_resource| true)
+        resources_state.pkgutil_modules_infos(py, prefix, state.optimize_level, |resource| {
+            _PathEntryFinder::is_visible("", &resource.name)
+        })
     }
 
     // Canonicalize the path to the current executable or raise an OSError.
@@ -1025,7 +1027,7 @@ py_class!(class _PathEntryFinder |py| {
         self.finder(py).call_method(py, "invalidate_caches", NoArgs, None)
     }
 
-    def iter_modules(&self, prefix: Option<&str> = None) -> PyResult<PyList> {
+    def iter_modules(&self, prefix: &str = "") -> PyResult<PyList> {
         self.iter_modules_impl(py, prefix)
     }
 
@@ -1073,11 +1075,11 @@ impl _PathEntryFinder {
             .map(|spec| if spec == py.None() { None } else { Some(spec) })
     }
 
-    fn iter_modules_impl(&self, py: Python, prefix: Option<&str>) -> PyResult<PyList> {
+    fn iter_modules_impl(&self, py: Python, prefix: &str) -> PyResult<PyList> {
         let state = self.finder(py).cast_as::<OxidizedFinder>(py)?.state(py);
         let modules = state.get_resources_state().pkgutil_modules_infos(
             py,
-            prefix.map(|p| p.to_string()),
+            Some(prefix.to_string()),
             state.optimize_level,
             |resource| _PathEntryFinder::is_visible(self.package(py), &resource.name),
         );
--- a/pyembed/src/importer.rs
+++ b/pyembed/src/importer.rs
@@ -533,6 +533,10 @@ py_class!(class OxidizedFinder |py| {
         oxidized_finder_new(py, relative_path_origin)
     }
 
+    def path_hook(&self, path: PyObject) -> PyResult<_PathEntryFinder> {
+        self.path_hook_impl(py, path)
+    }
+
     def index_bytes(&self, data: PyObject) -> PyResult<PyObject> {
         self.index_bytes_impl(py, data)
     }
@@ -905,7 +909,212 @@ impl OxidizedFinder {
             None
         };
 
-        resources_state.pkgutil_modules_infos(py, prefix, state.optimize_level)
+        resources_state.pkgutil_modules_infos(py, prefix, state.optimize_level, |_resource| true)
+    }
+
+    fn path_hook_impl(&self, py: Python, path: PyObject) -> PyResult<_PathEntryFinder> {
+        // Compute _PathEntryFinder::package
+        let pkg = {
+            // Get the canonicalized path to the current executable or raise an OSError.
+            let current_exe = {
+                let current_exe = &self.state(py).get_resources_state().current_exe;
+                current_exe.canonicalize().map_err(|err| {
+                    let exe = current_exe.display();
+                    // Use a Python expression instead of PyErr::new to ensure we get the
+                    // right OSError subclass.
+                    let strerror = format!("cannot open current executable: '{}'", exe);
+                    let raw_os_error = err
+                        .raw_os_error()
+                        .map_or_else(|| "None".to_string(), |err_code| err_code.to_string());
+                    let code = if cfg!(windows) {
+                        format!(
+                            "OSError(None, r\"{}\", r\"{}\", {})",
+                            strerror, exe, raw_os_error
+                        )
+                    } else {
+                        format!("OSError({}, r\"{}\", r\"{}\")", raw_os_error, strerror, exe)
+                    };
+                    py.eval(&code, None, None).map_or_else(
+                        |err_creating_exc| err_creating_exc,
+                        |exc| PyErr::from_instance(py, exc),
+                    )
+                })
+            }?;
+            let not_exe_err = || {
+                let msg = format!(
+                    "path {} does not begin in \'{}\' (if you're sure it does, check Python's file-system encoding)", &path, current_exe.display());
+                let mut err = PyErr::new::<ImportError, _>(py, msg);
+                let exc = err.instance(py);
+                if let Err(setattr_err) = exc.setattr(py, "path", path.clone_ref(py)) {
+                    setattr_err
+                } else {
+                    err
+                }
+            };
+            let pbuf = pyobject_to_pathbuf(py, path.clone_ref(py))?;
+            let mut p: &std::path::Path = pbuf.as_ref();
+            // path could point "inside" current_exe. If we can't canonicalize,
+            // strip the last component and check again. (zipimporter does
+            // somethings similar)
+            let mut abs_p = p.canonicalize();
+            // Count the number of components at the end of `path` will be in pkg
+            let mut parts: usize = 0;
+            while abs_p.is_err() {
+                p = match p.parent() {
+                    Some(parent) => {
+                        parts += 1;
+                        parent
+                    }
+                    None => {
+                        return Err(not_exe_err());
+                    }
+                };
+                abs_p = p.canonicalize();
+            }
+            let abs_p = abs_p.unwrap();
+            if abs_p != current_exe {
+                return Err(not_exe_err());
+            }
+            // Extract the part of `path` after current_exe
+            // FIXME: This takes two allocations, but it shouldn't even use one.
+            let tail = pbuf
+                .iter()
+                .rev()
+                .take(parts)
+                .collect::<std::path::PathBuf>()
+                .iter()
+                .rev()
+                .collect::<std::path::PathBuf>();
+            _PathEntryFinder::parse_path_to_pkg(py, &tail)?
+        };
+        let finder = OxidizedFinder::create_instance(py, Arc::clone(self.state(py)))?;
+        // sys.path_hooks is only read after importlib._bootstrap_internal has
+        // been setup, either in that module or after initialization by
+        // PyImport_GetImporter checking for the existence of config->run_filename
+        // has been setup, so it's safe to import os now.
+        let paths = PyList::new(py, &[py.import("os")?.call(py, "fspath", (path,), None)?]);
+        _PathEntryFinder::create_instance(py, finder.into_object(), paths, pkg)
+    }
+}
+
+// A (mostly compliant) `importlib.abc.PathEntryFinder` that delegates paths
+// within the current executable to the `OxidizedFinder` whose `path_hook`
+// method created it. This is a private implementation detail.
+py_class!(class _PathEntryFinder |py| {
+    // A `importlib.abc.MetaPathFinder`, presumably but not necessarily an
+    // `OxidizedFinder`. However, `_PathEntryFinder::iter_modules` will raise
+    // a `TypeError` when called if `finder` is not an `OxidizedFinder`.
+    data finder: PyObject;
+    // A list containing a single either str or bytes, normalized from the `path`
+    // argument to `OxidizedFinder::path_hook`.
+    data path: PyList;
+    // If the entry of `path` is `os.path.join(sys.executable, "pkg", "mod")`,
+    // then `package` is `"pkg.mod"`.
+    data package: String;
+
+    def find_spec(&self, fullname: &str, target: Option<PyModule> = None) -> PyResult<Option<PyObject>> {
+        self.find_spec_impl(py, fullname, target)
+    }
+
+    def invalidate_caches(&self) -> PyResult<PyObject> {
+        self.finder(py).call_method(py, "invalidate_caches", NoArgs, None)
+    }
+
+    def iter_modules(&self, prefix: Option<&str> = None) -> PyResult<PyList> {
+        self.iter_modules_impl(py, prefix)
+    }
+
+    // Private getter. Just for testing.
+    @property def _package(&self) -> PyResult<String> {
+        Ok(self.package(py).clone())
+    }
+});
+
+impl _PathEntryFinder {
+    // `name` is considered _visible_ in `package` if `name` is in the first
+    // level of the package. The computation is purely textual. See this
+    // module's `test_path_entry_finder::test_is_visible` test.
+    fn is_visible(package: &str, name: &str) -> bool {
+        if package.starts_with('.') || package.ends_with('.') || package.contains("..") {
+            return false;
+        }
+        // This implementation can be simplified when str::rsplit_once stabilize
+        // https://doc.rust-lang.org/std/primitive.str.html#method.rsplit_once
+        let mut split = name.rsplitn(2, '.');
+        match split.next() {
+            Some(module) if !module.is_empty() && !module.contains('.') => {}
+            _ => {
+                return false;
+            }
+        }
+        match split.next() {
+            Some(pkg) if !package.is_empty() && pkg == package => true,
+            None if package.is_empty() => true,
+            _ => false,
+        }
+    }
+
+    fn find_spec_impl(
+        &self,
+        py: Python,
+        fullname: &str,
+        target: Option<PyModule>,
+    ) -> PyResult<Option<PyObject>> {
+        if !_PathEntryFinder::is_visible(self.package(py), fullname) {
+            return Ok(Some(py.None()));
+        }
+        self.finder(py)
+            .call_method(py, "find_spec", (fullname, self.path(py), target), None)
+            .map(|spec| if spec == py.None() { None } else { Some(spec) })
+    }
+
+    fn iter_modules_impl(&self, py: Python, prefix: Option<&str>) -> PyResult<PyList> {
+        let state = self.finder(py).cast_as::<OxidizedFinder>(py)?.state(py);
+        let modules = state.get_resources_state().pkgutil_modules_infos(
+            py,
+            prefix.map(|p| p.to_string()),
+            state.optimize_level,
+            |resource| _PathEntryFinder::is_visible(self.package(py), &resource.name),
+        );
+        // unwrap() is safe because pkgutil_modules_infos returns a PyList cast
+        // into a PyObject.
+        Ok(modules?.cast_into(py).unwrap())
+    }
+
+    // Transform b"pkg/mod" into "pkg.mod" on behalf of `OxidizedFinder::path_hook`.
+    // Raise an `ImportError` if decoding `path` fails.
+    fn parse_path_to_pkg(py: Python, path: &std::path::Path) -> PyResult<String> {
+        let mut pkg = Vec::new();
+        for part in path.iter() {
+            pkg.push(
+                // Use Python's filesystem encoding to ensure correctly
+                // round-tripping the str Python package name from the
+                // potentially non-Unicode path.
+                // https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_170
+                // https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+                crate::conversion::path_to_pyobject(py, std::path::Path::new(part))?
+                    .cast_as::<PyString>(py)?
+                    .to_string(py)
+                    .map_err(|mut unicode_err| {
+                        // Need to raise an ImportError if can't decode. For
+                        // clarity, we also attach the UnicodeDecodeError.
+                        // https://docs.python.org/3/reference/import.html#path-entry-finders
+                        let mut imp_err = PyErr::new::<ImportError, _>(py, "cannot decode path");
+                        let imp_exc = imp_err.instance(py);
+                        if let Err(err) = imp_exc.setattr(py, "__suppress_context__", true) {
+                            err
+                        } else if let Err(err) =
+                            imp_exc.setattr(py, "__cause__", unicode_err.instance(py))
+                        {
+                            err
+                        } else {
+                            imp_err
+                        }
+                    })?
+                    .into_owned(),
+            );
+        }
+        Ok(pkg.join("."))
     }
 }
 
@@ -1466,17 +1675,16 @@ fn module_init(py: Python, m: &PyModule) -> PyResult<()> {
     Ok(())
 }
 
-/// Replace all meta path importers with an OxidizedFinder instance.
+/// Replace all meta path importers with an OxidizedFinder instance and return it.
 ///
 /// This is called after PyInit_* to finish the initialization of the
-/// module. Its state struct is updated. A new instance of the meta path
-/// importer is constructed and registered on sys.meta_path.
+/// module. Its state struct is updated.
 #[cfg(not(library_mode = "extension"))]
 pub(crate) fn replace_meta_path_importers<'a>(
     py: Python,
     m: &PyModule,
     resources_state: Box<PythonResourcesState<'a, u8>>,
-) -> PyResult<()> {
+) -> PyResult<PyObject> {
     let mut state = get_module_state(py, m)?;
 
     let sys_module = py.import("sys")?;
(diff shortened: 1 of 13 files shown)


wkschwartz:
I couldn't run tests on Windows because of #349.

wkschwartz:
I see there are merge conflicts now. I’ll rebase into master in the next couple hours.

wkschwartz:
@indygreg I rebased onto master and fixed the merge conflicts. This is ready for review.

wkschwartz:
@indygreg Does this PR have a shot at being merged? I understand if it'll take awhile, of course. But if it's not a good fit for PyOxidizer, I'll need to make some backup plans for some projects I'm working on.

indygreg:
> @indygreg Does this PR have a shot at being merged? I understand if it'll take awhile, of course. But if it's not a good fit for PyOxidizer, I'll need to make some backup plans for some projects I'm working on.

It does. But it's a sizeable PR and I've been busy with other things in the past few weeks. It's on the short list of things to look at the next time I spend considerable time hacking on PyOxidizer.

wkschwartz:
Thank you! I totally understand it taking awhile, and of course let me know if I can do anything to simplify the review, answer questions, correct anything, etc.

wkschwartz:
I turned off “Allow edits and access to secrets by maintainers”. Of course I still welcome your commits to the PR (via a separate PR). I just need to make sure a force-push doesn’t orphan 357bbc3 (the current tip of the branch). I’ve asked my team to install directly from it.

indygreg:
Well, this PR is next on my list of PRs to review and merge. But due to its size and complexity (I need to grok how `__path__` filtering is even supposed to work), it might take a while. I already grabbed a beer to wind down on Friday. So it almost certainly won't merge tonight :)

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+                    strerror, exe, raw_os_error
+                )
+            } else {
+                format!("OSError({}, r\"{}\", r\"{}\")", raw_os_error, strerror, exe)
+            };
+            py.eval(&code, None, None).map_or_else(
+                |err_creating_exc| err_creating_exc,
+                |exc| PyErr::from_instance(py, exc),
+            )
+        })
```
indygreg, line 938: I just merged another PR to introduce `dunce::canonicalize()` to the `pyembed` crate. We should definitely use that crate to perform the path canonicalization so UNC paths don't enter the picture on Windows. `.current_exe` here may already be normalized, but I'd have to double check the code to be sure.

What I'm trying to say is that all this code can likely be deleted. At the worst we'd introduce another call to `dunce::canonicalize()`. Ideally `.current_exe` is already normalized on storage.

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+                } else {
+                    err
+                }
+            };
+            let pbuf = pyobject_to_pathbuf(py, path.clone_ref(py))?;
+            let mut p: &std::path::Path = pbuf.as_ref();
+            // path could point "inside" current_exe. If we can't canonicalize,
+            // strip the last component and check again. (zipimporter does
+            // somethings similar)
+            let mut abs_p = p.canonicalize();
```
indygreg, line 961: This should use `dunce::canonicalize()` to avoid UNC paths on Windows.

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+                    Some(parent) => {
+                        parts += 1;
+                        parent
+                    }
+                    None => {
+                        return Err(not_exe_err());
+                    }
+                };
+                abs_p = p.canonicalize();
+            }
```
indygreg, line 975: We can likely use `Path.components()` plus iterators here to do something more elegantly. Although this code isn't horrible. Another strategy would be to use https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix to remove the exe part of the path. e.g. `abs_p.strip_prefix(current_exe).map_err(...).components().iter().count()` (or something like that - we may need to filter the components as part of the iterator bit).

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+            // Extract the part of `path` after current_exe
+            // FIXME: This takes two allocations, but it shouldn't even use one.
+            let tail = pbuf
+                .iter()
+                .rev()
+                .take(parts)
+                .collect::<std::path::PathBuf>()
+                .iter()
+                .rev()
+                .collect::<std::path::PathBuf>();
```
indygreg, line 989: Yeah, we want `Path.strip_prefix()` here :) But I like that you have a strong command on iterators: that's impressive for someone newish to Rust!

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+            _PathEntryFinder::parse_path_to_pkg(py, &tail)?
+        };
+        let finder = OxidizedFinder::create_instance(py, Arc::clone(self.state(py)))?;
```
indygreg, line 992: Nit: I'm pretty sure `self.state(py).clone()` will work here.

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+        // sys.path_hooks is only read after importlib._bootstrap_internal has
+        // been setup, either in that module or after initialization by
+        // PyImport_GetImporter checking for the existence of config->run_filename
+        // has been setup, so it's safe to import os now.
+        let paths = PyList::new(py, &[py.import("os")?.call(py, "fspath", (path,), None)?]);
```
indygreg, line 997: Since this is normalizing an argument value and it should be safe to call at the top of the function, could you please move this there?

While you are here, possibly add a comment saying why this is necessary (I know the reason, but it wasn't obvious when I first saw this code).

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+        _PathEntryFinder::create_instance(py, finder.into_object(), paths, pkg)
+    }
+}
+
+// A (mostly compliant) `importlib.abc.PathEntryFinder` that delegates paths
+// within the current executable to the `OxidizedFinder` whose `path_hook`
+// method created it. This is a private implementation detail.
+py_class!(class _PathEntryFinder |py| {
```
indygreg, line 1005: Why is this prefixed with `_`? Perhaps we could rename it `OxidizedPathEntryFinder` so the name is consistent with other types defined by this Python extension?

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+        }
+    }
+
+    fn find_spec_impl(
+        &self,
+        py: Python,
+        fullname: &str,
+        target: Option<PyModule>,
+    ) -> PyResult<Option<PyObject>> {
+        if !_PathEntryFinder::is_visible(self.package(py), fullname) {
```
indygreg, line 1065: Can we use `Self::is_visible()` here to avoid typing the type name?

pyembed/src/importer.rs
```diff
@@ -900,7 +904,219 @@ impl OxidizedFinder {
...
+                            imp_exc.setattr(py, "__cause__", unicode_err.instance(py))
+                        {
+                            err
+                        } else {
+                            imp_err
+                        }
+                    })?
+                    .into_owned(),
+            );
+        }
```
indygreg, line 1118: I'll have to think about this a bit.

We know that `/` and `\` are the only valid directory separators. So assuming we get a relative path, we could potentially extract the `&Path` view we need, convert that to a `PyObject`, and encode. This feels much simpler than iterating over path parts and encoding each.

Also, I'd prefer if any path encoding code existed in the `conversion` module. Getting that correct is difficult and it is generic code that IMO should live in its own function.

indygreg (commented):
I haven't looked at the tests in extreme detail, so this is only a partial review.

I want to commend you for the documentation changes and the details in the PR and linked issue: it made reviewing/understanding the new functionality a lot easier than I was anticipating!

The Rust code overall looks pretty solid! There are some style/convention nits, but nothing too major. I didn't see anything too major to give me concern over this feature and anticipate it being merged after another iteration or two.

I left a number of comments, but they were mostly minor.

There's a lot of changes in here that could be performed in isolation. If you want to factor those into their own commits / PRs, we could start merging things sooner. But the size of this isn't too bad: it's mostly your super comprehensive tests blowing up the size! Those tests will definitely come in handy when some of the path code is refactored per review comments.

wkschwartz:
Fantastic, thank you! My responses to your review may dribble in over the weekend, but I may not get to this in earnest until early next week.

pyembed/src/importer.rs
wkschwartz: I'll change the name if you want, but allow me to explain my choices first.

I prefixed the name with `_` to mark `oxidized_importer._PathEntryFinder` as [private](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) from Python (none of the `py_class!`es in the module are [`pub`lic](http://dgrunwald.github.io/rust-cpython/doc/cpython/macro.py_class.html#syntax) from Rust). It's always possible to make a private API public later, but it's hard to change or remove a public API. My habit is to start off a feature with the smallest possible public API surface area. As my users and I gain experience with the feature, I add more public interfaces.

I eschewed prefixing the name with `Oxidized` for two reasons: (1) `_PathEntryFinder` is currently private, so Python code (other than pyembed's own tests) never needs to touch it; and (2) I generally prefer to [avoid stutter](https://blog.golang.org/package-names#TOC_3.) when naming interfaces.

Of course, like I said, I'll change the name however you want.

pyembed/src/importer.rs
wkschwartz: ## Simplifications

Taking your suggestion of assuming that all possible path separators are `/` and `\` and converting the `path: &Path` to a `PyObject` in one go, I came up with the an implementation (see diff below the fold) that differs mostly in that it no longer takes advantage of `std::path::Iter`'s already knowing what the path separators are. Now, the new code *is* slightly shorter, but it hard codes `/` and `\` and iterates through the resulting `String` twice on Windows instead of once. Maybe using `str.replace` requires less re-allocation of memory than `Vec.push`?

<details><summary>Diff showing new implementation taking your advice</summary>

```diff
diff --git a/pyembed/src/importer.rs b/pyembed/src/importer.rs
index 2325480a..194a14ef 100644
--- a/pyembed/src/importer.rs
+++ b/pyembed/src/importer.rs
@@ -1098,37 +1098,34 @@ impl _PathEntryFinder {
     // Transform b"pkg/mod" into "pkg.mod" on behalf of `OxidizedFinder::path_hook`.
     // Raise an `ImportError` if decoding `path` fails.
     fn parse_path_to_pkg(py: Python, path: &std::path::Path) -> PyResult<String> {
-        let mut pkg = Vec::new();
-        for part in path.iter() {
-            pkg.push(
-                // Use Python's filesystem encoding to ensure correctly
-                // round-tripping the str Python package name from the
-                // potentially non-Unicode path.
-                // https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_170
-                // https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
-                crate::conversion::path_to_pyobject(py, std::path::Path::new(part))?
-                    .cast_as::<PyString>(py)?
-                    .to_string(py)
-                    .map_err(|mut unicode_err| {
-                        // Need to raise an ImportError if can't decode. For
-                        // clarity, we also attach the UnicodeDecodeError.
-                        // https://docs.python.org/3/reference/import.html#path-entry-finders
-                        let mut imp_err = PyErr::new::<ImportError, _>(py, "cannot decode path");
-                        let imp_exc = imp_err.instance(py);
-                        if let Err(err) = imp_exc.setattr(py, "__suppress_context__", true) {
-                            err
-                        } else if let Err(err) =
-                            imp_exc.setattr(py, "__cause__", unicode_err.instance(py))
-                        {
-                            err
-                        } else {
-                            imp_err
-                        }
-                    })?
-                    .into_owned(),
-            );
-        }
-        Ok(pkg.join("."))
+        // Use Python's filesystem encoding to ensure correctly
+        // round-tripping the str Python package name from the
+        // potentially non-Unicode path.
+        // https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_170
+        // https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+        let pkg = crate::conversion::path_to_pyobject(py, path)?
+            .cast_as::<PyString>(py)?
+            .to_string(py)
+            .map_err(|mut unicode_err| {
+                // Need to raise an ImportError if can't decode. For
+                // clarity, we also attach the UnicodeDecodeError.
+                // https://docs.python.org/3/reference/import.html#path-entry-finders
+                let mut imp_err = PyErr::new::<ImportError, _>(py, "cannot decode path");
+                let imp_exc = imp_err.instance(py);
+                if let Err(err) = imp_exc.setattr(py, "__suppress_context__", true) {
+                    err
+                } else if let Err(err) =
+                    imp_exc.setattr(py, "__cause__", unicode_err.instance(py))
+                {
+                    err
+                } else {
+                    imp_err
+                }
+            })?
+            .replace("/", ".");
+        #[cfg(windows)]
+        let pkg = pkg.replace("\\", ".");
+        Ok(pkg)
     }
 }

 ```

</details>

## Extracting something to the `conversion` module

It's not clear to me what part of the function to extract into the `conversion` module. Sixteen of the lines (including comments and braces) of this function in both the old and new implementations do nothing but raise an [`ImportError`] from a `UnicodeDecodeError`. This "conversion" is highly [specific to path hooks](https://docs.python.org/3/reference/import.html#path-entry-finders):
> The hook should expect either a string or bytes object; the encoding of bytes objects is up to the hook (e.g. it may be a file system encoding, UTF-8, or something else), and if the hook cannot decode the argument, it should raise [`ImportError`].

The import documentation makes no other mention of text encodings. To me, this means raising an [`ImportError`] from a `UnicodeDecodeError` need not happen anywhere else in the code base.

In the new implementation, this leaves seven other lines that aren't comments, one of which returns from the function, and three of which replace `/` and `\` with `.`.

The only other thing I can think to change in the `conversion` module is to change the return type of `path_to_pyobject` to be [`PyResult`](http://dgrunwald.github.io/rust-cpython/doc/cpython/type.PyResult.html)`<`[`PyString`](http://dgrunwald.github.io/rust-cpython/doc/cpython/struct.PyString.html)`>` to avoid the `.cast_as::<PyString>(py)?` line in `parse_path_to_pkg`. Had I written `path_to_pyobject` from scratch, maybe* that's the return type I would have picked to begin with? But such a change seems out of scope for this PR. I'm happy to make the change if you disagree.

<details><summary>*</summary>

Not necessarily. I could see an argument for going with `PyBytes` over `PyString`. But again, this strikes me as a debate for a different issue or PR.

</details>

[`ImportError`]: https://docs.python.org/3/library/exceptions.html#ImportError

File: pyembed/src/importer.rs (hunk 5 above)
Comment (wkschwartz) [reply pyembed/src/importer.rs:992]: I'm going to get rid of the line altogether and instead pass `self.as_object().clone_ref(py)` as the `finder` argument below.

**Comment by wkschwartz** [comment]
I'm about half done with your review, but won't be able to finish it until tomorrow or the next day. I'm also a little concerned what will happen to 357bbc3 (which I've asked my team to use) if I force push with my local changes (which includes rebasing onto `main`). Maybe I need to open up a dummy PR with 357bbc3 to make sure it doesn't get garbage collected? Maybe it doesn't matter.

File: pyembed/src/importer.rs (hunk 1 above)
Comment (wkschwartz) [reply pyembed/src/importer.rs:938]: > `.current_exe` here may already be normalized, but I'd have to double check the code to be sure.

[`dunce::canonicalize`] sets `ResolvedOxidizedPythonInterpreterConfig.current_exe` only if `OxidizedPythonInterpreterConfig.exe` wasn't already set and `OxidizedPythonInterpreterConfig::resolve` was the constructor:
https://github.com/indygreg/PyOxidizer/blob/05f4a67fe436728ee9d70d30768e1aa4c520ab9c/pyembed/src/config.rs#L282-L291

This might hold for [`MainPythonInterpreter::new`](https://github.com/indygreg/PyOxidizer/blob/05f4a67fe436728ee9d70d30768e1aa4c520ab9c/pyembed/src/interpreter.rs#L69) (again, if `config.exe` isn't set), but will not hold for `OxidizedImporter`s created from Python:
https://github.com/indygreg/PyOxidizer/blob/05f4a67fe436728ee9d70d30768e1aa4c520ab9c/pyembed/src/importer.rs#L949
which calls
https://github.com/indygreg/PyOxidizer/blob/05f4a67fe436728ee9d70d30768e1aa4c520ab9c/pyembed/src/python_resources.rs#L444

-----

> What I'm trying to say is that all this code can likely be deleted... Ideally `.current_exe` is already normalized on storage.

The ideal doesn't seem currently to match the code. I am eager to avoid bugs in whatever we merge into `main` from this PR, so I'd like to push back against deleting `exe_realpath` _for now_.

----

> At the worst we'd introduce another call to `dunce::canonicalize()`.

The surgery required to ensure [all locations that set `current_exe`](https://github.com/indygreg/PyOxidizer/blob/05f4a67fe436728ee9d70d30768e1aa4c520ab9c/pyembed/src/python_resources.rs#L395) are properly normalized strikes me as being beyond the scope of this PR. I count at least two different structs that own a `current_exe` (`OxidizedPythonInterpreterConfig` and `PythonResourcesState`), and both have a variety of constructors, not all of which appear to be used. Without narrowing down the places where `current_exe` gets set, we risk a bad or missing canonicalization sneaking back in in the future.

Instead I have two alternative proposals. I'd be okay with either one, so please let me know what you'd want to do.

1. Shore up the current code as is and merge it for now. I would create a ticket for fixing the `current_exe` canonicalization situation and the ticket would mention the opportunity for removing `OxidizedImporter::exe_realpath`.
2. Change the code to check for `os.join(sys.executable, "pkg", "mod")`. This would better match the documentation as currently written and would be easier for Python code to verify. It would differ from the way that embedded packages' `__path__`s are created, which is to prefix with `current_exe`. Note that `current_exe` does not have to be the same as `sys.executable`. In particular, `sys.executable` is writable from Python. Other reasons `current_exe` and `sys.executable` may not match include race conditions in reading symlinks during interpreter startup.

Until I hear back from you, I'm going to carry forward with shoring up the code as is. In particular, there's a bug in how the `OSError` is constructed that I'll fix.

[`dunce::canonicalize`]: https://docs.rs/dunce/1.0.1/dunce/fn.canonicalize.html

File: pyembed/src/importer.rs (hunk 2 above)
Comment (wkschwartz) [reply pyembed/src/importer.rs:961]: Because we check `abs_p == current_exe` later on, what matters is that both are canonicalized using the same function: either both with `dunce` or both with `std`. As discussed above in https://github.com/indygreg/PyOxidizer/pull/343#discussion_r585777519, whether we keep `exe_realpath` or just use `self.state(py).get_resources_state().current_exe` directly, `current_exe` will not necessarily have been canonicalized by `dunce`. So switching this line to using `dunce` would cause `path_hook` to raise `ImportError`s when it should have returned a `_PathEntryFinder`.

Note that _if we keep `exe_realpath`_, then it doesn't matter at all whether we canonicalize using `dunce` or `std`. The canonicalized form is used only to check that `current_exe` matches the head of `abs_p`. `dunce::canonicalize` calls `std::fs::canonicalize` and then does some extra processing that won't be observable. The performance difference surely doesn't matter, but my bias is generally to avoid extra dependencies where they aren't needed.

File: pyembed/src/importer.rs (hunk 3 above)
Comment (wkschwartz) [reply pyembed/src/importer.rs:975]: I had considered [`Path::strip_prefix`] here and below, but here's a bug that could arise from that. Suppose `current_exe` is `/path/to/myapp` but `sys.executable` is `/symlink/to/myapp` and that both are correct (`sys.executable` often points to a symlink when Python is running inside a virtual environment).
```python
pkgutil.iter_modules([os.path.join(sys.executable, 'importlib')])
```
should list all the modules in the `importlib` package, but will instead be empty because `/symlink/to/myapp/importlib` does not start with `/path/to/myapp`. (Maybe we only care that `pkgutil.iter_modules(importlib.__path__)` works, which it would without canonicalization. But that seems to me to be a confusing API.) Calling `canonicalize` on `/symlink/to/myapp/importlib` will fail because that path doesn't exist. So we either need all the paths that point at `/path/to/myapp` (impossible) and use  [`Path::strip_prefix`] until one works (slow), or we need to peel the end of the path off one component at a time like I do here or like [`zipimport` does](https://github.com/python/cpython/blob/b36349a647b2bf8174f0e736a4fc347e92ae204e/Lib/zipimport.py#L74-L90).

I am now realizing I don't have this as a test case, so I'll add it.

[`Path::strip_prefix`]: https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix

pyembed/src/importer.rs
wkschwartz: Not sure why I didn't see it before: I can replace this with `pbuf.strip_prefix(p).unwrap()`. I've got a local patch that does this (and some other simplifications now that I don't need `parts`). Whole function is 20 lines shorter.

pyembed/src/importer.rs
wkschwartz: That `OSError` construction bug is now fixed in 97096f1a03d49572c2077ce42f0f390369dae7e9

pyembed/src/importer.rs
wkschwartz: dbdd169d972674603cad53a75e3ba9b555054a88

pyembed/src/importer.rs
wkschwartz: 3ae1efb2677b3e01ef3f2fbe6784e2059c017680

pyembed/src/importer.rs
wkschwartz: 3da60fe800f5e28a2b5493fe99de5622473852b0

pyembed/src/importer.rs
wkschwartz: 7e2411bbc7640898fd9dbd97dc9fb9ad7156545e

pyembed/src/importer.rs
wkschwartz: I added the symlink test case in 9cbaf7c3a38817eda0e9a1e88cb4c598d5e2051a but I haven't run it on Windows yet where I used a directory junction since symlinks are security cordoned there

wkschwartz:
I think I've done everything I can until I hear back from you on comments I've left above. It would help me if you could hit the "Resolve conversation" button your feedback that you feel that I have adequately addressed.

> There's a lot of changes in here that could be performed in isolation. If you want to factor those into their own commits / PRs, we could start merging things sooner. But the size of this isn't too bad: it's mostly your super comprehensive tests blowing up the size! 

I'm not really sure how to break this stuff up. In any case, I think it'll be easier to test against my use case if everything is in one PR—but I can still do some rebasing to group changes into more logical commits if that'll help you review. For now I'll leave everything as is. That way you can see how I've responded to the review for this round.

pyembed/src/importer.rs
wkschwartz: Oh, I should mention one more thing here: The `raise ImportError() from unicode_decode_error` code path is currently untested because [the test](https://github.com/indygreg/PyOxidizer/pull/343/files#diff-8be82d8159d7c11fdb6e5fe044a1092032f151c0c8ff57e1bf4dff3babc64324R275-R284) for it fails with a `SystemError` due to dgrunwald/rust-cpython#246. That bug has now been patched but won't be available at least until the next release of rust-cpython.

pyembed/src/importer.rs
indygreg: If the type gets exposed to Python code, IMO it should be exported from the module to facilitate `isinstance()` checks and other type sniffing.

pyembed/src/importer.rs
indygreg: I'll have to think about what we should do with the path hooks conversion. Path encoding is hard. Python gets some of it wrong. I'd prefer to not do that with PyOxidizer. But I have the feeling that compatibility with Python is probably more important here.

Anyway, I like the new proposed code in your comment above. I'll think about it more during the next review. And if it is something that I want to change, I can do that in flight or as a follow-up: I don't want you to worry too much about nuanced issue.

pyembed/src/importer.rs
wkschwartz: Done. 8625a80d283d5c698d677854058bb1eec4b6509d

pyembed/src/importer.rs
wkschwartz: Personally, I'm happy with the new code. Frankly, I'm not sure there's much more we can do with it [until we can test it](https://github.com/indygreg/PyOxidizer/pull/343#discussion_r585920650).

indygreg (approved):
I'm going to land this with an in-flight change to rename the new class to `OxidizedPathEntryFinder`.

Sorry this PR took so long to review. I got side-tracked with various things.

There may be some test failures as fallout from this change. I'll deal with this if they produce in CI.

indygreg:
@wkschwartz When I sat down to implement `pkg_resources` support, I found myself asking questions about the implementation of path entry finder, notably the path canonicalization and normalization pieces. I found myself changing behavior to be simpler and less brittle at run-time. I just wanted to give you a heads up that I'm about to push a significant refactor of some of this code to the `main` branch. I think the changes are for the better. But it is clear from my work on this that you had (still have?) a stronger grasp on path hooks than I did (do?).

Anyway, I wanted to give you a courtesy heads up about the inbound changes to a) give you the opportunity to inspect them for unwanted changes b) reassure you that your original work was solid and the amount of code churn I performed should not negatively reflect on your work.

Thank you again for contributing this work! It unblocked implementing `pkg_resources` integration, which has long been on my TODO list!

wkschwartz:
It looks like the patch in https://github.com/indygreg/PyOxidizer/pull/343#issuecomment-810512857 didn't make into main. If you're seeing test failures on windows, take a look at the fixes there.

> give you the opportunity to inspect them for unwanted changes

I'm not seeing a PR for the changes you mentioned. Would you please list the commits I should look at?

PR #344, opened 2020-12-20
#345: pyembed: BUG: SIGSEGV crash in safe code

wkschwartz:
This PR adds a test for the segmentation fault reported in #345. If you remove the `compile_fail` annotation from the test and run `cargo test --doc` from the `pyembed` directory on macOS (and, I'm guessing, on other platforms), you'll get

```
---- src/interpreter.rs - interpreter::MainPythonInterpreter::acquire_gil (line 427) stdout ----
Test executable failed (terminated by signal).
```

If you move the test code from a doc test to a regular test, you'll get a little more output and see that that's a `SIGSEGV` signal. Isn't Rust supposed to prevent memory errors in code that doesn't have `unsafe {}` around it?

I *think* what's happening is that the lifetime bounds on `pyembed::MainPythonInterpreter` aren't strict enough to prevent the example from compiling in the first place, and it triggers a use-after-free error in the unsafe code that acquires and frees Python's GIL.

I'm submitting this as a draft pull request for now since I don't have a solution to this bug.

### DIFF
--- a/pyembed/src/interpreter.rs
+++ b/pyembed/src/interpreter.rs
@@ -422,6 +422,18 @@ impl<'python, 'interpreter, 'resources> MainPythonInterpreter<'python, 'interpre
     }
 
     /// Ensure the Python GIL is acquired, returning a handle on the interpreter.
+    /// The lifetime bounds prevent use-after-free errors like the following,
+    /// which does not compile.
+    /// ```compile_fail
+    /// let py = {
+    ///     let mut config = pyembed::OxidizedPythonInterpreterConfig::default();
+    ///     config.interpreter_config.parse_argv = Some(false);
+    ///     config.set_missing_path_configuration = false;
+    ///     let mut interp = pyembed::MainPythonInterpreter::new(config).unwrap();
+    ///     interp.acquire_gil().unwrap()
+    /// };
+    /// py.import("sys").unwrap();
+    /// ```
     pub fn acquire_gil(&mut self) -> Result<Python<'python>, &'static str> {
         match self.interpreter_state {
             InterpreterState::NotStarted => {


wkschwartz:
Maybe adding `lifetime: `[`std::marker::PhantomData`]`<&'interpreter ()>,` to the `MainPythonInterpreter` struct would help? The datatype being `()` seems wrong though.

[`std::marker::PhantomData`]: https://doc.rust-lang.org/std/marker/struct.PhantomData.html#unused-lifetime-parameters

indygreg:
Thanks for reporting this. It is likely a lifetime annotation failure, coupled with use of `unsafe`. Although it could be an issue with CPython's APIs not being _safe_: I've run into a handful of issues where calling certain CPython APIs in certain order triggers segfaults in C land. Lots of undocumented behavior quirks in CPython...

Regarding your comment about memory errors and `unsafe`, Rust can only guarantee memory safety if `unsafe` isn't used or all code using `unsafe` is sound and properly lifetime annotated. If there is a bug in the code inside `unsafe`, Rust code outside an `unsafe {}` can crash due to memory or logic issues. If you are dealing with a C API that is prone to crashing, even the most risk adverse `unsafe {}` code could crash due to bugs inside the C code.

indygreg:
I'm going to close this PR, as I've locally implemented the test as a `#[test]`, outside of a code comment. We can continue tracking this bug in #344.

Thank you again for contributing the test case!

wkschwartz:
The reason I put mine in a comment—which is a doc test and is run by `cargo test --doc`—is because I would expect that once the bug is fixed, the test would fail to compile, which is what [the `compile_fail` line](https://github.com/indygreg/PyOxidizer/pull/344/files#diff-ef5e9dd97e7fe4231ff2ca0d5173282e07f4418997eebf8fa9fa6a7342b7cd7eR427) specifically tests for (see [reference](https://doc.rust-lang.org/edition-guide/rust-2018/rustdoc/documentation-tests-can-now-compile-fail.html)). Anyway, if you figured out how to write a regular `#[test]` for it, that'll work too.

PR #351, opened 2020-12-30
pyembed: API: MainPythonInterpreter ownership prevents runtime errors

wkschwartz:
Previously, the `acquire_gil` method of a `MainPythonInterpreter` would return an `Err` if called after `py_runmain`. Coordinating this required a state type `InterpreterState`, which `MainPythonInterpreter` managed manually.

Now `py_runmain` takes ownership of its receiver, so Rust will refuse even to compile a program that attempts to call `acquire_gil` after `py_runmain`.

Because such an error is now impossible, `acquire_gil` has become infallible. I removed the `Result` wrapping of the `Python` return type to reflect this.

These are breaking changes, but `py_runmain`'s taking ownership of `MainPythonInterpreter` is safer, taking advantage of the benefits of Rust that make dealing with its borrow checker worthwhile. And if `acquire_gil` continued
to return a `Result`, it would be confusing for beginners, and the constant `.unwrap()`ing would be un-ergonomic for maintainers.

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -55,6 +55,8 @@ Backwards Compatibility Notes
 * ``show_alloc_count`` has been removed from types representing Python
   interpreter configuration because support for this feature was removed in
   Python 3.9.
+* ``pyembed::MainPythonInterpreter.acquire_gil()``'s signature has changed, now
+  returning a ``Python`` value directly without wrapping it in a ``Result``.
 
 Bug Fixes
 ^^^^^^^^^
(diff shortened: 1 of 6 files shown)


indygreg (approved):
I like the change to the API and this proposed change looks good to me! Thank you for contributing it.

I'll merge this once CI passes. (GitHub Actions seems to be struggling with macOS runner availability today.)

indygreg:
I cherry picked this commit and pushed as e86b2f46ed6b449bdb912900b0ac83576ad5ebe9. Sadly, GitHub doesn't allow me to mark the PR as merged. But it is in fact part of the `main` branch now.

Thank you again for your contribution!

PR #354, opened 2021-01-06
Use absolute exe path in pyembed.OxidizedPythonInterpreterConfig

gi0baro:
SSIA

This allows to keep correct paths for filesystem importer in case of symbolic linking built executables on all operative systems, as `std::env::current_exe` [might not return consistent results](https://doc.rust-lang.org/std/env/fn.current_exe.html#platform-specific-behavior).

### DIFF
--- a/pyembed/Cargo.toml
+++ b/pyembed/Cargo.toml
@@ -17,6 +17,7 @@ links = "pythonXY"
 # Update documentation in lib.rs when new dependencies are added.
 anyhow = "1.0"
 cpython = "0.5.2"
+dunce = "1.0.1"
 jemalloc-sys = { version = "0.3", optional = true }
 lazy_static = "1.4"
 libc = "0.2"
--- a/pyembed/deny.toml
+++ b/pyembed/deny.toml
@@ -5,6 +5,7 @@
 unlicensed = "deny"
 allow = [
     "0BSD",
+    "CC0-1.0",
     "MIT",
     "MPL-2.0",
     "Python-2.0",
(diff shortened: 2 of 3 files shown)


indygreg (changes requested):
Thank you for the contribution.

I wholeheartedly support the intent of this change.

However, there's an issue with this PR.

`canonicalize()` will yield a UNC path on Windows. If this path makes its way to CPython, it will likely have unintended consequences and break things. So we need to ensure we use a more traditional Windows path instead.

Unfortunately, Rust's standard library doesn't have a way to get a non-UNC symlink resolved absolute path. There are crates like `path-absolutize` and `path-dedot` that we could potentially use here. Although I'm unsure if they cover this exact case. So we may have to write a little bit extra code.

Also, if you want to add a test that the resolved path is not UNC, that would be appreciated. But it isn't required: this is something I can implement myself after this merges.

gi0baro:
@indygreg I switched to [dunce](https://lib.rs/crates/dunce) which should quickly fix the Windows UNC thing.

CI is not passing on nightly, but it seems to be unrelated to this change.
Regarding tests: this is the early beginning of my Rust journey, then you can probably write such tests quickly and even properly compared to me.

gi0baro:
@indygreg any chance to see this merged?

indygreg:
> @indygreg any chance to see this merged?

I took a break from PyOxidizer for several weeks. I'm back to development and am slowly making my through open PRs. I should hopefully have feedback on this anywhere from a few minutes from now to a few days. I have to sort out dependencies between PRs and figure out which ones I want to tackle first.

indygreg (approved):
This looks great! Will push this to `main` shortly.

Thank you for your contribution!

PR #369, opened 2021-02-26
change Lazy_static to once_cell, no macros, faster compilation

ryancinsight:
Instead of using Lazy static I converted to once_cell, std library version is only on nightly so need to use crate for 1.45 compatibility. Once_cell does not use macros and it is more flexible, I stuck to sync for single threading. 

from async_std forum once_cell is faster than Lazy_static but macros in general appear to be  known to compile slower.
10:04:54|~/projects/once_cell|pl-tests✓
λ cargo run --release --example bench_vs_lazy_static
   Compiling once_cell v1.2.0 (/home/matklad/projects/once_cell)
    Finished release [optimized] target(s) in 0.39s
     Running `target/release/examples/bench_vs_lazy_static`
once_cell:   486.122531ms (total time of many accesses accesses)
lazy_static: 579.513044ms


ryancinsight:
I'm not sure I know the difference between stable/ beta windows vs nightly/no label windows for half to fail at downloading vcredist...does stable and beta windows not download vc_redist in tests?

ryancinsight:
BTW, How invested are you in wix installation btw? I'm not sure installing vcredist for you will solve all problems so I'm looking into making a dependency walker that maps the exe, dll, and pyd files and figures out what other dll are need. some of my swigged libraries actually have tbb and openmp dlls which aren't in vcredist for example. furthermore, some packages include duplicates of vc_runtime which really messes with me and is typically useless in that packages folder. Then, I'm modifying warp-packer to use zstandard and tar.ztd the install folder with proper dll's into an exe that checks version installled and updates if older than new exe. The install location can be user set, though I prefer hidden folders, and the executable becomes a symlink to hidden install folder. It's only 2 mb currently but I'm working to reduce the dependencies right now: https://github.com/ryancinsight/Zpack. Dependency walking recursively has been a little more complex though...

indygreg:
> I'm not sure I know the difference between stable/ beta windows vs nightly/no label windows for half to fail at downloading vcredist...does stable and beta windows not download vc_redist in tests?

Those failures are likely a race condition in our code for downloading files. It attempts to use file locks to ensure at most 1 writer is downloading. The code is pretty janky and is likely buggy.

> BTW, How invested are you in wix installation btw?

I want PyOxidizer (actually _tugger_) to support as many distributable formats as possible. I started with WiX for MSI files because it is the MSI builder I'm most familiar with. It is also super powerful and used by a few projects I'm associated with that want to use PyOxidizer. So it made sense to start there.

> I'm not sure installing vcredist for you will solve all problems so I'm looking into making a dependency walker that maps the exe, dll, and pyd files and figures out what other dll are need.

You are correct: ultimately we'll need to examine binaries and look for dependencies. This is on my not-yet-written-down roadmap. At least some of that code would belong in the `tugger-binary-analysis` crate. The VC++ Redistributable is necessary for dynamically linked Windows distributions. So we should absolutely handle installing it (either a DLL next to the built binary or via its `.exe` installer).

> Then, I'm modifying warp-packer to use zstandard and tar.ztd the install folder with proper dll's into an exe that checks version installled and updates if older than new exe. The install location can be user set, though I prefer hidden folders, and the executable becomes a symlink to hidden install folder. It's only 2 mb currently but I'm working to reduce the dependencies right now: https://github.com/ryancinsight/Zpack. Dependency walking recursively has been a little more complex though...

That's a cool project worth watching!

One of the properties of many other Python application distribution formats is that they self-extract to a temporary directory or FUSE filesystem and run Python code from that. `pyembed` is certainly capable of doing this, but it isn't yet implemented. I would very much like to evolve the Python packed resources format so that it is both compressed (likely with zstd) and supports extracting to the filesystem at run-time. This of course can be done generically, which is I guess what you are attempting to do with zpack. If the code is easily reusable, it is definitely something I'd consider integrating into `pyembed` / PyOxidizer someday!

indygreg (approved):
This patch looks good!

I'll be applying some minor fixups locally (mainly to run `rustfmt` on all changed files). (I should really get a CI job to verify `rustfmt` is happy so people know the repo has a policy of only checking in code formatted with `rustfmt`.)

PR #375, opened 2021-03-18
rcodesign: add support for custom entitlements

aspenluxxxy:
Adds an optional `-e` argument, which takes a path to an entitlements plist.


indygreg (approved):
Nice! Thank you for the contribution.

I was planning to add some missing features to the CLI today and this was on the list. This PR saves me a bit of work. But be forewarned I may rework the argument names later today as more features are implemented and the interface needs to evolve to facilitate end-user ergonomics. Nothing against your excellent work here: this PR is perfect.

PR #377, opened 2021-03-19
Update default configuration template

B3QL:
Update configuration accordingly to latest documentation.
https://pyoxidizer.readthedocs.io/en/latest/config_type_python_packaging_policy.html#extension-module-filter


indygreg (approved):
Good catch. Thank you for the contribution!

PR #385, opened 2021-04-24
Distribute pyoxidizer binary via PyPI

messense:
This PR adds a GitHub Actions workflow to build and publish pyoxidizer binary wheels to PyPI so that it can be installed with just `pip install pyoxidizer`.

Feel free to close this if you don't want to do it.

TODO:

- [ ] cargo build caching


messense:
Here is an example run on my fork: https://github.com/messense/PyOxidizer/actions/runs/780241171

indygreg:
Thanks for providing this PR! I'm definitely supportive of lowering the barrier to end-users installing PyOxidizer!

Now let's talk about issues.

First, the automatic publishing to PyPI is a non-starter for me. I don't trust the security of GitHub Actions with credentials to perform releases. I understand I may be in the minority here and that many projects do publish releases direct from GitHub. But GitHub Actions makes it way too easy to leak credentials and the consequences could be disastrous. I insist that a human person (me) be in the driver's seat to publishing a release and I'm moving to a release workflow where this will require physical access to a hardware token.

This shouldn't be a big deal: it just changes the release mechanics a little. I'm definitely in favor of generating wheels from CI, so this PR will land in some form.

Second, I'm unsure how publishing a wheel with a Rust executable (+ wheel metadata files) to PyPI will be received. I'm in favor of using PyPI and common Python packaging tools as a distribution channel/vector for PyOxidizer. I just don't know if [ab]using PyPI as a distribution channel for a Rust application will be tolerated. If it is, great: we'll release to PyPI! I'll try to solicit feedback outside this PR. If you have other examples of non-Python applications doing this, let me know. I do know the topic of PyPI for application distribution has come up before in the Python community: I just don't know what the latest thoughts are.

Third, while the tools used in the PR to produce wheels are likely fine to use for now, I wanted to give you a courtesy heads up that it is likely that PyOxidizer will someday support producing wheels natively. This PR made me realize that I'm kind of surprised we don't already support it! (I guess nobody has asked for that feature.) PyOxidizer already has enough primitives in its internals to support doing this: they just aren't all glued together to make it turnkey. Using PyOxidizer to self-build its own wheels is probably in the future. But it is most definitely scope bloat for this PR and shouldn't prevent a working solution from landing. Perfect is the enemy of good.

Fourth, I'm slowly adding pre-built binaries of PyOxidizer to the releases. Since the ultimate goal here seems to be *enable people to run PyOxidizer quickly*, I'm curious if other pre-built binary solutions will be sufficient? The existing GitHub Actions CI currently has support for generating the following release artifacts:

* A statically linked x86-64 Linux executable
* Universal macOS DMG
* Windows x86 and x86-64 MSIs and a multi-arch exe installer

We could just as easily build/publish standalone executables for macOS and Windows as well if those would be helpful.

Finally. I generally support the idea of building wheels containing a `pyoxidizer` executable and want to see this PR land in some form. But if we want it to be possible to `pip install pyoxidizer` from PyPI, I'd like to get some form of approval before I upload releases to PyPI.

indygreg:
The CI in the `main` branch now produces the new following artifacts:

* x86 and x86-64 `pyoxidizer.exe` CLI executables
* x86-64 and aarch64 `pyoxidizer` macOS executables
* x86-64 + aarch64 `pyoxidizer` universal/fat macOS executable

I plan to publish these in some form in the next release and they will be available in the GitHub release as artifacts.

Just thought you'd like to know in case it helps with your use case.

messense:
> If you have other examples of non-Python applications doing this, let me know. I do know the topic of PyPI for application distribution has come up before in the Python community: I just don't know what the latest thoughts are.

Some examples I know:

* [py-spy](https://pypi.org/project/py-spy/): https://github.com/benfred/py-spy#how-are-you-distributing-rust-executable-binaries-over-pypi
* [cmake](https://pypi.org/project/cmake/): https://github.com/scikit-build/cmake-python-distributions
* [ninja](https://pypi.org/project/ninja/): https://github.com/scikit-build/ninja-python-distributions

messense:
> The CI in the `main` branch now produces the new following artifacts:
> 
> * x86 and x86-64 `pyoxidizer.exe` CLI executables
> * x86-64 and aarch64 `pyoxidizer` macOS executables
> * x86-64 + aarch64 `pyoxidizer` universal/fat macOS executable
> 
> I plan to publish these in some form in the next release and they will be available in the GitHub release as artifacts.
> 
> Just thought you'd like to know in case it helps with your use case.

👍 It helps, just not as simple as `pip install pyoxidizer`, and needs to hardcode download urls for each platform.

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+  pull_request:
+
+jobs:
+  macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
```
indygreg, line 14: Please change this to Python 3.9 (assuming it works). I like to keep all my automation current and Python 3.6 is nearly end of lifed.

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,104 @@
+name: PyPI Release
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  macos:
+    runs-on: macos-latest
```
indygreg, line 9: Please change this to `10.15`: I like to keep CI deterministic.

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+          python-version: 3.6
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: aarch64-apple-darwin
+          profile: minimal
+          default: true
+      - name: Install maturin
+        run: pip install maturin
```
indygreg, line 23: I prefer to have all dependencies pinned using the `pip-compile --generate-hashes` tool from `pip-tools` and then use `pip install -r requirements.txt`.

Feel free to do this in this PR. Or I'll do it as a follow-up.

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          maturin build -m pyoxidizer/Cargo.toml -b bin --target x86_64-apple-darwin --release --out dist --no-sdist
+          pip install pyoxidizer --no-index --find-links dist --force-reinstall
+      - name: Build wheels - universal2
+        env:
+          DEVELOPER_DIR: /Applications/Xcode.app/Contents/Developer
+          MACOSX_DEPLOYMENT_TARGET: '10.9'
+        run: |
+          # set SDKROOT for C dependencies like bzip2
+          export SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
```
indygreg, line 34: Is this needed? The macOS machines in CI typically pick this up automatically.

If it is needed, I highly prefer to pin the location to the absolute path so SDKs don't change out from under us. Something like `export SDKROOT=/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk`.

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+          maturin build -m pyoxidizer/Cargo.toml -b bin --release --universal2 --out dist --no-sdist
+          pip install pyoxidizer --no-index --find-links dist --force-reinstall
+      - name: Upload wheels
+        uses: actions/upload-artifact@v2
+        with:
+          name: wheels
+          path: dist
+
+  windows:
+    runs-on: windows-latest
```
indygreg, line 44: Please use `windows-2019` for determinism.

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
```
indygreg, line 11: This needs

```
with:
  fetch-depth: 0
```

in order to avoid a wonky run-time bug. tl;dr PyOxidizer's build script looks at the Git history of its checkout to bake some settings into the built binary. It gets confused when running from shallow clones, which GitHub Actions does by default.

This change needs to be made throughout the file.

.github/workflows/release_pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+
+  linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [
+          { manylinux: '2010', target: "x86_64-unknown-linux-musl", image_tag: "x86_64-musl" },
+          { manylinux: '2010', target: "i686-unknown-linux-musl", image_tag: "i686-musl" },
+          { manylinux: '2014', target: "aarch64-unknown-linux-musl", image_tag: "aarch64-musl" },
+          { manylinux: '2014', target: "armv7-unknown-linux-musleabihf", image_tag: "armv7-musleabihf" },
```
indygreg, line 84: Do we need the 2014 variant? The year component in the `manylinux` platform tag mainly denotes symbol compatibility with glibc. Since we statically link with musl, we should run on all Linux, so the 2014 variant probably isn't needed.

indygreg (commented):
This looks pretty good!

I left a handful of comments. Most of them minor. Many of the comments apply to multiple places in the file. But I only called them out once.

Also, I'm trying to consolidate all the release artifact generation into `release.yml`. If you wouldn't mind merging these jobs into that workflow, it would be appreciated. Otherwise, I'll likely do it after this lands.

.github/workflows/release_pypi.yml
messense: It's for aarch64 and armv7 Linux which are only available since manylinux2014. https://www.python.org/dev/peps/pep-0599/

We can remove it if pyoxidizer doesn't support architectures other that x86_64/i686.

.github/workflows/release_pypi.yml
messense: Failed to get it work for both macOS, Windows and Linux so I've reverted back to just `pip install maturin`.

indygreg (approved):
Thank you for making all the changes!

I am going to fetch and squash this PR locally and hopefully push it to `main` sometime today.

I will likely perform some minor follow-ups to tweak behavior so things are slightly more deterministic and reproducible. But the crux of your work should remain intact.

I may or may not publish any or all of the wheels to PyPI during the next release: I would like to perform more QA before committing to doing so.

PR #397, opened 2021-04-28
libc unneeded for memory allocations, cc build now available for snmalloc 

ryancinsight:
Hello,

First set of changes are removal of unneeded calls to libc crate and additionally use of build_cc feature for snmalloc to remove need for cmake.


ryancinsight:
need to make changes, submit later

PR #398, opened 2021-04-28
libc unneeded for memory allocations, cc build now available for snmalloc

ryancinsight:
Might need changes but I think we can reduce libc requirements, reduced compilation time by 7 seconds and tests by 0.05 seconds when not using for snmalloc. Also added build_cc feature for snmalloc to no longer use cmake


pyembed/src/interpreter_config.rs
```diff
@@ -6,12 +6,13 @@
 
 use {
     crate::{config::ResolvedOxidizedPythonInterpreterConfig, NewInterpreterError},
-    libc::{c_int, size_t, wchar_t},
+    core::{usize,i32,u16},
```
indygreg, line 9: Clippy says `u16` is unused outside of Windows.

So we need to import any of these primitives anyway? They are part of the standard library prelude and should always exist.

pyembed/src/interpreter_config.rs
```diff
@@ -21,14 +22,18 @@ use {
 
 #[cfg(target_family = "unix")]
 use std::{ffi::NulError, os::unix::ffi::OsStrExt};
+#[cfg(target_family = "unix")]
+type WcharT = i32;
```
indygreg, line 26: No need to get fancy. Can we please keep this named as `wchar_t`? This will require an inline annotation to `#[allow(..)]` whatever it needs to be to silence the compiler warning about type names.

pyembed/src/python_resources.rs
```diff
@@ -948,6 +949,9 @@ impl<'a> PythonResourcesState<'a, u8> {
         // this functionality some day. But it should likely never be the default
         // because it goes against the spirit of requiring all resources to be
         // known ahead-of-time.
+
+        let enoent: c_int = 2;
```
indygreg, line 953: Please make this a `const`.

indygreg (commented):
This looks mostly good. Thanks for removing the extra explicit dependency! I didn't realize our surface area of utilization of that crate was so small.

That being said, the fact that the crate didn't disappear completely from `new-project-cargo.lock` tells me it is still used somewhere. So this does not actually remove the crate dependency. It looks to still be used by `cpython`, `jemallocator`, and others.

pyembed/src/interpreter_config.rs
ryancinsight: Very true, I'll remove the import statements

ryancinsight:
> This looks mostly good. Thanks for removing the extra explicit dependency! I didn't realize our surface area of utilization of that crate was so small.
> 
> That being said, the fact that the crate didn't disappear completely from `new-project-cargo.lock` tells me it is still used somewhere. So this does not actually remove the crate dependency. It looks to still be used by `cpython`, `jemallocator`, and others.

Unfortunately, it is still being used by others, in some situations it does make sense due to the various os having different definitions for ctypes, like I probably wont mess with strings. However, I may submit pull requests to the others to be more direct if possible. For snmalloc it made no sense to use, with cmake build it took it to 27 seconds with libc and 20 seconds without to compile. On the cc version, which I added to pyoxidizer build features, it is 16 seconds with libc and only 9-10 seconds without libc. On simple tests it was also saving fractions of a second consistently so I still think being as direct as possible instead of going ctype->libc crate -> rust type alias ->libc crate -> rust type alias is still better.

pyembed/src/python_resources.rs
ryancinsight: Done

ryancinsight:
I was thinking of adding the memory allocator feature options soon as well, I just added windows 7/8 compatibility and use of c++20 standards to snmalloc-rs that I'm hoping to add here as I have already run into issues with transferring binaries to teammates using windows 7.

Do you have suggestions on implementation? I think renaming with designation of features could work but may also be too verbose unless only adding a couple features:

[dependencies]
foo = "0.1"
foo_*= { version = "0.1", registry= "custom1", package = "foo", features=* }
foo_** = { version = "0.1", registry = "custom2", package = "foo", features=** }

indygreg:
OK, you've convinced me that shedding the explicit `libc` dependency is worthwhile. I didn't realize its build time was so long!

Since you may send patches to other crates using it, I do intend to port `pyembed` to `PyO3` at some time. When, I'm not sure. But this does seem inevitable given the better ergonomics of that crate. It has a somewhat larger dependency tree.

Regarding the extra Cargo features, I'd prefer to avoid the complexity, if possible. But I don't believe Cargo allows you to overrule the features of indirect dependency crates, right? e.g. when we compile our Python executable, we can't have its `Cargo.toml`  override the features that `pyembed` will use when compiling e.g. `snmalloc-rs`, right? What all this means is that `pyembed` would have to create different features mapping to different features in its direct dependencies, right? In that case, I suppose I have no choice but to say yes to adding more features to the `pyembed` crate.

indygreg (approved):
Thanks for the follow-ups to the review comments. I'm going to squash and rebase this locally then hopefully push to the `main` branch shortly. Although there's a bit of a CI backlog at the moment, so it might take an hour or so.

Issue #324, opened 2020-11-12
PyO3 support


davidhewitt:
A PyO3 maintainer here unfamiliar with PyOxidizer, just discovered it this afternoon. 👋 

Looks like a really cool project! We reasonably often have pain with users who want to embed Python in Rust programs; there's all kinds of subtleties around static linking etc. So it'd be great to understand a little bit about what would be necessary for PyOxidizer to work with PyO3 projects (if it doesn't already).

I'd be very happy to help out with any implementation work needed and also add some notes to the [PyO3 guide](https://pyo3.rs) / add an example in the PyO3 repo.

Also, while reading your docs I noticed this section: https://pyoxidizer.readthedocs.io/en/stable/status.html?highlight=pyo3#pyo3

> PyO3
>
> PyO3 are alternate Rust bindings to Python from rust-cpython, which is what pyembed currently uses.
>
> The PyO3 bindings seem to be ergonomically better than rust-cpython. PyOxidizer may switch to PyO3 someday.

I'd be very happy to answer any questions you have about using PyO3 in `pyembed` & help with the implementation if you're interested in switching. IMO helping with downstream usage from time to time gives me ideas how to improve the ergonomics in PyO3!

indygreg:
👋! 

It should be possible for binaries created with PyOxidizer to load extensions built against PyO3. But it all comes down to how things are linked. As I'm sure you know, this can be a thorny problem space. If extensions are built in a way that is compatible with PyPI supported binary wheels, things should _just work_. But if we have to recompile from source to get static linking, I expect this to be a minefield. (The Python extension module building ecosystem doesn't really account for the possibility of statically linked extensions, for better or worse.)

PyOxidizer's run-time component (the `pyembed` crate) currently uses the `cpython` + `python3-sys` crates. I would **love** to move these to PyO3 at some point. I just bumped the minimum Rust version requirement, unblocking us from adopting PyO3.

I was actually planning to spend a few hours looking at things and open an issue against PyO3 listing all the changes we'd need. But you seem to have beat me to it :)

[...]

PR #415, opened 2021-06-25
`spdx=0.4.1` dependency increases MSRV to 1.52

EndilWayfare:
Running `cargo install pyoxidizer` with Rust 1.50.0 on Windows 10 failed, complaining about use of unstable feature `str_split_once`.

The usage occurs in `spdx::lexer::find_document_and_license_ref`. The change was introduced in `spdx@0.4.1` as part of removing dependence on `regex`/`lazy_static`. That's awesome for compilation-speed/binary-size, but now all depending crates transitively depend on Rust >=1.52, the version which stabilized `str::split_once`. This isn't a big deal for me personally, because I don't have a reason to use an older toolchain (I just hadn't upgraded for a bit). But, since this project's MSRV is 1.46, I thought this might become an issue for someone else.

`python-packaging`, `tugger-licensing`, and `tugger-licensing-net` all depend on `spdx = "0.4"`. Cloning the repo and building from source succeeds; the Cargo.lock pins `spdx=0.4.0`. However, `cargo install --path pyoxidizer` fails just like `cargo install pyoxidizer`. Apparently, `cargo install` doesn't respect `Cargo.lock` unless you use `--locked`.

Constraining `spdx = "= 0.4.0"` fixes this, at the cost of shackling users with Rust >=1.52 to the old implementation. On the other hand, that may be desirable for consistency between MSRV up to current version.

Alternatively, the documentation could add a suggestion to use `cargo install --locked` for older toolchains.

### DIFF
--- a/python-packaging/Cargo.toml
+++ b/python-packaging/Cargo.toml
@@ -20,7 +20,7 @@ mailparse = "0.13"
 once_cell = "1.7"
 regex = "1"
 sha2 = { version = "0.9", optional = true }
-spdx = "0.4"
+spdx = "= 0.4.0"
 time = { version = "0.1", optional = true }
 walkdir = "2"
 zip = { version = "0.5", optional = true }
--- a/tugger-licensing-net/Cargo.toml
+++ b/tugger-licensing-net/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 [dependencies]
 anyhow = "1.0"
 reqwest = { version = "0.11", default-features = false, features = ["blocking", "rustls-tls"] }
-spdx = "0.4"
+spdx = "= 0.4.0"
 url = "2.2"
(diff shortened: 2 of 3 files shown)


EndilWayfare:
Ok, I just found out that `pyoxidizer` enforces Rust 1.52.1 as the minimum rust version for use with the cli.

I only thought of the MSRV because I had seen #324, being curious as to the relationship of this project with `pyo3`. There, it was specified as 1.46, and I found 1.46 associated with the most recent usage of "minimum" searching in [Project History](https://pyoxidizer.readthedocs.io/en/latest/pyoxidizer_history.html). I see now that 1.52.1 is mentioned much more recently in the context of "Manged Rust toolchain".

I think this PR is entirely useless, then. :sweat_smile:

PR #422, opened 2021-07-19
Support for overriding URLs via environment variables

martinvonz:
I haven't been able to test this yet (I don't have a Windows machine). Is there a good place to add automated tests for this?

### DIFF
--- a/tugger-common/src/http.rs
+++ b/tugger-common/src/http.rs
@@ -7,12 +7,14 @@ use {
     fs2::FileExt,
     sha2::Digest,
     slog::warn,
-    std::{io::Read, path::Path},
+    std::{fs::File, io::Read, path::Path},
     url::Url,
 };
 
 /// Defines remote content that can be downloaded securely.
 pub struct RemoteContent {
+    /// Identifier used for configuring an override for the URL.
+    pub name: String,
     pub url: String,
     pub sha256: String,
 }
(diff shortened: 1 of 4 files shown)


tugger-common/src/http.rs
```diff
@@ -10,9 +10,12 @@ use {
     std::{io::Read, path::Path},
     url::Url,
 };
+use std::fs::File;
```
indygreg, line 13: Nit: please move this into the statement above.

indygreg (changes requested):
Please rebase this on top of `main`, as I think the Cargo changes won't be necessary since I mass updated dependencies earlier today.

And do you have any more context for this PR? I'm guessing this has something to do with a firewall in a hermetic build environment?

As for testing this, there are a few integration tests that validate building Windows installers work. That code exercises the downloading of these files. We could add one of these environment variables to the CI job to exercise this functionality. But I'm OK not having explicit testing of this feature.

martinvonz:
> Please rebase this on top of `main`, as I think the Cargo changes won't be necessary since I mass updated dependencies earlier today.

Done.

> And do you have any more context for this PR? I'm guessing this has something to do with a firewall in a hermetic build environment?

Yes, that's correct (see the commit message). We download all dependencies before the build so the build can run without internet access.

> As for testing this, there are a few integration tests that validate building Windows installers work. That code exercises the downloading of these files. We could add one of these environment variables to the CI job to exercise this functionality. But I'm OK not having explicit testing of this feature.

I'm also OK with not having explicit testing :)

PR #428, opened 2021-08-06
make apple crate config optional on windows

ryancinsight:
I downloaded most recent version of pyoxidizer and have been trying to install but ran into crates for apple potions of code failing on windows, can we please make the apple portion config optional on windows, this is minimum I had to do to work on windows as a start.

### DIFF
--- a/apple-bundle/Cargo.toml
+++ b/apple-bundle/Cargo.toml
@@ -10,11 +10,11 @@ homepage = "https://github.com/indygreg/PyOxidizer"
 repository = "https://github.com/indygreg/PyOxidizer.git"
 readme = "README.md"
 
-[dependencies]
+[target.'cfg(target_vender="apple")'.dependencies]
 anyhow = "1.0"
 plist = "1.2"
 walkdir = "2.3"
 
-[dependencies.tugger-file-manifest]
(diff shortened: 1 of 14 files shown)


ryancinsight:
@indygreg this was minimum working example to compile without apple crates on windows, though looks like your CI is more stringent than my laptop. Ill look at other configurations and better ways of doing when I get a chance but was hoping you would see this and reduce number of scripts actually determining build rather than using cfg options. its quite considerable the number of packages on windows alone and I can reduce by almost 100 cfg masking apple crates

indygreg:
What's the exact build problem you are trying to fix?

ryancinsight:
With Rust 1.53 I was needing to change some of the features on the crates currently being used to allow compatibility on older windows 7 laptops. Specifically with snmalloc in this case with win8compat feature. This requires generating a new lock file but I couldn't inability to find compatible crates with AES crate, tugged apple, crypto crates. So to generate a new lock file I had to cfg mask crates related to Apple which I'm not sure we need to compile on windows in the first place. It's a considerable number of packages on windows, if I remember correctly it went from 600 to 501 crates being compiled.

PR #433, opened 2021-08-11
Switch to PyO3

indygreg:
I'm opening this PR so we have a better place to look at the code diff.

Current state is I'm able to build a working binary with PyO3 on Linux. macOS probably works as well. Windows will need more work (and likely some changes to PyO3's build system).

There's run-time crash with release builds on Linux. I think we were tickling an existing bug, however, as the crash is due to memory safety related to `argv` handling during Python interpreter initialization. It is a heisenbug and goes away if I sprinkle print statements around the suspect code (which I was doing because the stack trace was nearly worthless since a lot of functions were inlined). I may have to break out `rr` to debug it.

Anyway, the build system integration is the piece of this PR that still needs the most work. The Rust code speaking to PyO3 in the `pyembed` crate should be pretty solid and ready for review. All tests pass against the PyO3 branch I'm running. At present time, that branch only has commits for which there are open PRs in the PyO3 project.

CC @davidhewitt if you feel like looking at how I'm using PyO3. The main think I know I'm not doing right is I haven't implemented a GC visitor for types that contain references to other Python types. But that was a preexisting bug I think.


davidhewitt:
👍 I will gladly take a good read of this, I might not get around to it until next week though just FYI.

indygreg:
I still need to get the `oxidized-importer` CI working. But basic PyOxidizer workflows appear to be working on this branch on Linux, Windows, and macOS!

This still needs a bit of cleanup before merge. And I'd like to get all the PyO3 changes this depends on landed in the upstream repo. But things are looking pretty good!

One unfortunate side-effect of the switchover is PyOxidizer's Rust projects now take longer to build. Only ~1.5s slower on my Ryzen 5950X. But in terms of CPU time we went from ~111s to ~165s. The slowdown is noticeable when running tests, which build many Rust projects as part of integration testing. `sccache` helps a bit. But we don't get many cache hits in tests due to use of temp directories. This newfound slowness might be a good excuse to improve test efficiency...

davidhewitt:
> One unfortunate side-effect of the switchover is PyOxidizer's Rust projects now take longer to build. Only ~1.5s slower on my Ryzen 5950X. But in terms of CPU time we went from ~111s to ~165s. The slowdown is noticeable when running tests, which build many Rust projects as part of integration testing. sccache helps a bit. But we don't get many cache hits in tests due to use of temp directories. This newfound slowness might be a good excuse to improve test efficiency...

Hmmm interesting. This is presumably due to using PyO3's proc macros instead of macro_rules? I wonder if there's much we can do to optimize them.

Also we had contributors previously who reduced llvm line counts (& compile times) by tweaking bits of the pyO3 codebase, I wonder if there's more that can be done there?

davidhewitt:
(Just a heads up that my offer to read this is still standing, however I'm giving a talk on PyO3 (virtually) at the Rust Manchester meetup this week, so I'm a bit swamped thinking about that until then. Hopefully at a quiet point over next weekend 🤞)

pyembed/build.rs
```diff
@@ -135,12 +136,23 @@ fn main() {
...
     println!("cargo:rustc-cfg=library_mode=\"{}\"", library_mode);
 
+    let pyo3_config_path = env::var("DEP_PYTHON_INTERPRETER_CONFIG_PATH")
+        .expect("DEP_PYTHON_INTERPRETER_CONFIG_PATH not found; this should have been set by the build system via the pyo3 crate");
+
+    let interpreter_config = PyO3InterpreterConfig::from_reader(
+        &std::fs::File::open(&pyo3_config_path)
+            .expect("could not open pyo3 config file for reading"),
+    )
+    .expect("failed to parse PyO3 configuration file");
```
davidhewitt, line 146: I'd be really curious to know if things work if you try this instead:

```suggestion
    let interpreter_config = pyo3_build_config::get();
```

... my expectation is that it should work fine; you use the `PYO3_CONFIG_FILE` env var so as long as that's present throughout the whole compilation, I expect this API should just work without needing any `DEP_` variable.

pyembed/src/conversion.rs
```diff
@@ -30,11 +37,11 @@ const SURROGATEESCAPE: &[u8] = b"surrogateescape\0";
 /// The optional encoding is the name of a Python encoding. If not used,
 /// the default system encoding will be used.
 #[cfg(target_family = "unix")]
-pub fn cpython_osstr_to_pyobject(
-    py: cpython::Python,
+pub fn osstr_to_pyobject<'p>(
```
davidhewitt, line 40: PyO3 has conversions from `OsStr` and `Path` to Python objects. These may potentially be usable for you instead of the ffi you're doing in this file?

pyembed/src/extension.rs
```diff
@@ -73,31 +66,29 @@ pub(crate) fn get_module_state<'a>(
...
 ///
-/// We don't use the macros in the cpython crate because they are somewhat
+/// We don't use the macros in the pyo3 crate because they are somewhat
 /// opinionated about how things should work. e.g. they call
 /// PyEval_InitThreads(), which is undesired. We want total control.
 #[allow(non_snake_case)]
-pub extern "C" fn PyInit_oxidized_importer() -> *mut oldpyffi::PyObject {
-    let py = unsafe { cpython::Python::assume_gil_acquired() };
+pub extern "C" fn PyInit_oxidized_importer() -> *mut pyffi::PyObject {
+    let py = unsafe { Python::assume_gil_acquired() };
```
davidhewitt, line 74: I actually removed the call to `PyEval_InitThreads` in https://github.com/PyO3/pyo3/pull/1630

... so you might find you're able to use the `#[pymodule]` proc macro fine?!

pyembed/src/extension.rs
```diff
@@ -143,114 +135,43 @@ fn register_pkg_resources(py: cpython::Python) -> cpython::PyResult<cpython::PyO
...
     let sys_module = py.import("sys")?;
-    let version_info = sys_module.get(py, "version_info")?;
+    let version_info = sys_module.getattr("version_info")?;
 
-    let major_version = version_info.getattr(py, "major")?.extract::<i32>(py)?;
-    let minor_version = version_info.getattr(py, "minor")?.extract::<i32>(py)?;
+    let major_version = version_info.getattr("major")?.extract::<i32>()?;
+    let minor_version = version_info.getattr("minor")?.extract::<i32>()?;
 
     if major_version < 3 || minor_version < 8 {
```
davidhewitt, line 149: You can use https://docs.rs/pyo3/0.14.2/pyo3/prelude/struct.Python.html#method.version_info here:

```suggestion
    py.version_info() <= (3, 8) {
```

pyembed/src/importer.rs
```diff
@@ -51,27 +57,26 @@ type py_init_fn = extern "C" fn() -> *mut oldpyffi::PyObject;
...
     if !existing_module.is_null() {
-        return Ok(unsafe { cpython::PyObject::from_owned_ptr(py, existing_module) });
+        return Ok(unsafe { PyObject::from_owned_ptr(py, existing_module) });
     }
 
     // An error occurred calling _PyImport_FindExtensionObjectEx(). Raise it.
-    if !unsafe { oldpyffi::PyErr_Occurred() }.is_null() {
-        return Err(cpython::PyErr::fetch(py));
+    if !unsafe { pyffi::PyErr_Occurred() }.is_null() {
+        return Err(PyErr::fetch(py).expect("Python exception should be set"));
```
davidhewitt, line 79: Interesting to see this. Makes me think `PyErr::api_call_failed` may want to be a public API of PyO3?

pyembed/src/importer.rs
```diff
@@ -282,99 +276,89 @@ pub(crate) struct ImporterState {
...
             1 => Ok(OptimizeLevel::One),
             2 => Ok(OptimizeLevel::Two),
-            _ => Err(cpython::PyErr::new::<cpython::exc::ValueError, _>(
-                py,
+            _ => Err(PyValueError::new_err(
                 "unexpected value for sys.flags.optimize",
             )),
         }?;
 
         let capsule = unsafe {
```
davidhewitt, line 348: I'd be interested in potentially working through https://github.com/PyO3/pyo3/issues/1193 sometime, looks like you might want it (and / or want to weigh in on its design).

pyembed/src/interpreter.rs
```diff
@@ -405,19 +427,12 @@ impl<'python, 'interpreter, 'resources> MainPythonInterpreter<'python, 'interpre
...
-            None => {
-                let gil = cpython::GILGuard::acquire();
-                let py = unsafe { cpython::Python::assume_gil_acquired() };
-
-                self._gil = Some(gil);
-                self.py = Some(py);
-
-                py
-            }
+    pub fn acquire_gil(&mut self) -> Python<'_> {
```
davidhewitt, line 430: I think the implementation of this function can just be `self.gil.get_or_insert_with(Python::acquire_gil).python()`

davidhewitt (commented):
This is an absolutely massive PR! I'm humbled by the amount of work you've put in here to try out PyO3 for PyOxidizer.

I've given this a brief read-through this evening and picked up on a few things that were interesting. There's kinda too much for me to read everything in full detail, however if there are particular bits you want to point me at I'm also happy to reread those?

indygreg:
I'm attempting to port this series to use the 0.14 branch. It _might_ be possible. But I won't know until I'm done.

I do know that `pyo3-build-config::InterpreterConfig` isn't a public symbol. Having it as a public symbol would save me from having to reinvent PyO3 config file writing logic.

indygreg:
OK. My local tests and CI say I have things working against 0.14.4!

The linking behavior can be extremely finicky though and getting it _just right_ has required days of effort in total. So I'm going to sit on this and keep refining the commits until I have more confidence this works and doesn't have undesirable regressions.

I think I can convince myself to land this without the additional `InterpreterConfig` fields requested in https://github.com/PyO3/pyo3/pull/1793 because the behavior is essentially the same as it was with `cpython`/`python3-sys`. However, said behavior is extremely hacky and brittle because essentially what I'm doing is having `python3-sys`/`pyo3` emit partial `cargo:` lines to control linking behavior and then I supplement the _missing_ pieces in `pyembed`'s build script. I would strongly prefer for `pyo3`'s build script to emit all the `cargo:` lines needed to link libpython. Because PyOxidizer's libpython is so different from normal Python distributions, it is better to have the generic `InterpreterConfig` extension mechanism rather than support my specific needs in the PyO3 project.

What I'm trying to say is I think I can merge this against 0.14.4. But I'd strongly prefer https://github.com/PyO3/pyo3/pull/1793 land so I can greatly simplify and de-risk PyOxidizer in the long term.

Also, I don't think I need the `links = ` `Cargo.toml` annotation either if I use `pyo3_build_config::get()`. (But I still think adding that annotation is a good idea.)

davidhewitt:
Brilliant!

It looks like I'll need to release a 0.14.5 anyway, so let's aim to merge the non-breaking bits of https://github.com/PyO3/pyo3/pull/1793 so that PyOxidizer is derisked 👍

indygreg:
OK. I've rebased onto the 0.14.5 release branch (commit 9dd16a828cedc583425088fb7221cffcc7058e2a) and things appear to be working great!

One small nit is the introduction of `pyo3-build-config` as a dependency in the `pyoxidizer` crate is requiring a Python interpreter be on `PATH`. This is a requirement that previously only existed for the `pyembed` crate (which holds all the Python API code). This dependency exists because `pyo3-build-config` has a build script that attempts to resolve a Python interpreter.

@davidhewitt: would you accept a patch to `pyo3-build-config` that adds a Cargo feature to effectively make the build script no-op? This would allow use of `pyo3-build-config` in _library mode_ and would help ensure the crate builds in more environments.

davidhewitt:
Awesome!

> @davidhewitt: would you accept a patch to `pyo3-build-config` that adds a Cargo feature to effectively make the build script no-op? This would allow use of `pyo3-build-config` in library mode and would help ensure the crate builds in more environments.

Yes, but I'd prefer the feature work slightly differently, for the sake of being additive. I think it could be called something like `resolve-config` or `embed-config` and should _enable_ the build script doing actual work, and maybe even add the `pyo3_build_config::get` API (and perhaps all the hidden exports used in `pyo3`'s build script too). If it was made a default feature (maybe additionally required by both `pyo3`  and `pyo3-macros-backend` for good measure), then it should be a non-breaking addition. I can then remove it from being a default feature in 0.15, because `maturin` would eventually like to use parts of `pyo3-build-config` too and won't need the config resolution either.

I think with that design, you'd be able to depend on `pyo3-build-config` with `default-features = false`? How does that sound?

indygreg:
> I think with that design, you'd be able to depend on `pyo3-build-config` with `default-features = false`? How does that sound?

This sounds reasonable. Let me to try to throw up a PR in the next few hours.

indygreg:
https://github.com/PyO3/pyo3/pull/1856

davidhewitt:
👍 that commit is now merged and also cherry-picked onto the `release-0.14.5` branch - do you want to give it another try? Fingers crossed we're all good for the PyO3 release now! 🤞

indygreg:
> 👍 that commit is now merged and also cherry-picked onto the `release-0.14.5` branch - do you want to give it another try? Fingers crossed we're all good for the PyO3 release now! 🤞

```
error: couldn't read /home/gps/src/pyoxidizer.git/target/debug/build/pyo3-build-config-412784ceff9b2afb/out/pyo3-build-config-file.txt: No such file or directory (os error 2)
  --> /home/gps/.cargo/git/checkouts/pyo3-a22e69bc62b9f0fd/c3796c4/pyo3-build-config/src/lib.rs:82:27
   |
82 | const CONFIG_FILE: &str = include_str!(concat!(env!("OUT_DIR"), "/pyo3-build-config-file.txt"));
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_str` (in Nightly builds, run with -Z macro-backtrace for more info)

error: couldn't read /home/gps/src/pyoxidizer.git/target/debug/build/pyo3-build-config-412784ceff9b2afb/out/pyo3-build-config-abi3.txt: No such file or directory (os error 2)
  --> /home/gps/.cargo/git/checkouts/pyo3-a22e69bc62b9f0fd/c3796c4/pyo3-build-config/src/lib.rs:87:27
   |
87 | const ABI3_CONFIG: &str = include_str!(concat!(env!("OUT_DIR"), "/pyo3-build-config-abi3.txt"));
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_str` (in Nightly builds, run with -Z macro-backtrace for more info)

error: couldn't read /home/gps/src/pyoxidizer.git/target/debug/build/pyo3-build-config-412784ceff9b2afb/out/pyo3-build-config.txt: No such file or directory (os error 2)
  --> /home/gps/.cargo/git/checkouts/pyo3-a22e69bc62b9f0fd/c3796c4/pyo3-build-config/src/lib.rs:92:27
   |
92 | const HOST_CONFIG: &str = include_str!(concat!(env!("OUT_DIR"), "/pyo3-build-config.txt"));
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_str` (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 3 previous errors

```

I guess we did need to conditionalize some of the variables and APIs after all. I'll get another PR up soon.

indygreg:
https://github.com/PyO3/pyo3/pull/1859 is up. I cherry picked this to the `release-0.14.5` branch and pushed to my fork as the `pyoxidizer` branch and rebased my series to use that branch. Everything looks good now.

davidhewitt:
Thanks for fixing. I've added that commit to the 0.14.5 release branch now. I'll put it live on the weekend, looks like we've made it through the initial challenges!

indygreg:
I was able to rebase on top of the 0.14.5 release and things look great on my end. I'm shoring up a few things before I merge this. But hopefully within the next 48 hours.

I'm super excited about this migration:

* The ergonomics of the Rust API make it vastly simpler for me to hack on the code interfacing with the Python interpreter.
* It unblocks the ability to cross-compile binaries since pyo3's crates don't need to run a `python` when provided a config file.
* It unblocks a lot of flexibility to move code around since we're no longer constrained by limitations in how `libpython` was hackily linked before. This should also lead to improved ergonomics around how builds work and making it easier for people to hack on Rust-based projects (as opposed to having PyOxidizer derive a Rust project for them).

So many great wins. Thank you, @davidhewitt, for your patience and willingness to work with me to get us to this point! I understand PyOxidizer is a bit of an edge case in terms of feature requests. I won't forget your support.

pyembed/src/extension.rs
indygreg: I'm not sure I can. That's because I have to define `PyModuleDef::m_size` to hold module state. I _might_ be able to get a reference to the struct somehow at run-time to populate it accordingly. But since I already have this code written, I'll probably keep it as is.

pyembed/src/conversion.rs
indygreg: Yes, it looks like I'll be able to port to your conversion implementations. I'll track this in a TODO and get around to it _sometime_.

indygreg:
This is now pushed to the `main` branch. So closing this PR.

Thanks again, @davidhewitt!

Issue #405, opened 2021-05-22
Is it possible to get fine-grained control over which stdlib extensions are included?


njsmith:
I was trying to figure out how to include specific stdlib extension modules (e.g. `_ssl`) but not others (e.g. `_asyncio`). I found:

- `PythonPackagingPolicy.extension_module_filter`: this lets you set a few general policies like `all` or `minimal`, but doesn't give control on a per-resource level AFAICT
- `PythonPackagingPolicy.register_resource_callback`: this gives you a ton of flexibility to configure most resources... but for stdlib extension modules, the `PythonExtensionModule` it receives seems to have `add_include=False` regardless of the setting of the `extension_module_filter`.

I think the most intuitive way would be for `extension_module_filter` to apply first and set the default values for `add_include` on all the stdlib modules, and then the resource callback would run second and have a chance to override that on a per-module basis. But apparently something else is going on?

Not sure if this is a documentation request or a feature request :-)

indygreg:
I looked at the source code and the registered resource callback _should_ be getting called for stdlib extension modules and its effects should be honored. So setting `.add_include = False` inside the callback should work.

However, there is some wonky Starlark code involved in the processing of resource callbacks and I wouldn't be surprised if mutations in the callbacks aren't working correctly. I'm also not sure if we have explicit test coverage of this scenario to catch failures.

I did refactor some Starlark code in the unreleased `main` branch a few weeks ago to hopefully minimize potential for copies of Starlark values to get out of sync with the original value (some Starlark values were effectively being copied instead of reference counted). This _may_ have fixed this apparent bug. But someone will need to verify.

Thanks for reporting this issue. It definitely feels like a bug to me. We definitely want resource callbacks to influence stdlib resources.

bbqsrc:
Can confirm the following worked on Windows:

```
def resource_callback(policy, resource):
    if type(resource) in ("PythonExtensionModule"):
        if resource.name == "_ctypes":
            resource.add_include = False
```

`extension_module_filter` does still run first though, and pre-filters everything out, so you can't toggle things back on.

bbqsrc:
Ahh, seems I spoke to soon. Linking still failed. Will update when I experiment more.

bbqsrc:
@indygreg I can confirm that the setter is called and the value is set to false, but it doesn't have much of an impact on the actual build behaviour.

WankkoRee:
The same problem still exists in the `0.24.0`. Is any solution?

PR #456, opened 2021-10-12
python-packaging: honor add_context.include for non-stdlib extensions

dae:
The setting was previously indiscriminately ignored, meaning it wasn't
possible to prevent third-party extension modules from being included
in the build.

Related: #405

### DIFF
--- a/pyoxidizer/docs/pyoxidizer_config_type_python_extension_module.rst
+++ b/pyoxidizer/docs/pyoxidizer_config_type_python_extension_module.rst
@@ -26,3 +26,6 @@
         (various)
 
         See :ref:`config_resource_add_attributes`.
+
+        Note: `add_include` is currently ignored for builtin modules, and can not
+        be used to control their inclusion.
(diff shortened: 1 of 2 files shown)


indygreg (approved):
Thanks for tracking this down.

I agree a partial solution here is better than none.

I'm going to rebase this and amend it slightly before pushing. I'll likely also tack on some tests in a follow-up commit because this behavior warrants testing!

indygreg:
As I was implementing tests, I found a few issues with how extension modules are handled. I'd like to get the tests shored up before landing this so I can be confident of what the behavior is. I may not get this landed tonight. But it is on my short list.

indygreg:
OK. So I've changed the API for the resource collector to return objects describing what actions were taken. I did this because as part of implementing tests for this, I encountered some weird behavior and found the existing logging around resource actions... lacking. So the new API enables us to print more useful log messages about what's going on.

This unfortunately introduced a merge conflict. The `return Ok(None);` in your PR should be changed to

```rust
return Ok((
    vec![AddResourceAction::NoInclude(extension_module.description())],
    None,
));
```

However, when I did this locally, a bunch of tests unrelated to the new ones started failing! I suspect the issue here has something to do with the original comment mentioning `built-in` extension modules and the new code only filtering on stdlib. The `built-in` set and the `stdlib` set aren't exactly the same and I think that is causing tests to fail.

This is a thorny situation. Having looked at this code, the tests I wrote, and the new log output (which appears to show duplicate extension adds), something is clearly off somewhere. I'll try to take a look at this again this week.

PR #460, opened 2021-10-15
starlark: Add PythonExecutable.write_aggregate_license_text to write aggregated license file

pajowu:
In audapolis (https://github.com/audapolis/audapolis) we use pyoxidizer to bundle a small webserver with an electron application. We want and need to list the licenses in the software and would like to do that automatically in the build process. The easiest way is to write a license acknowledgement-file during build. This PR implements a starlark method to do so.

This could be a first step to deal with the more general licensing issues when packaging pyoxidizer applications

Example:
```
exe = dist.to_python_executable(
      name="example",

      # If no argument passed, the default `PythonPackagingPolicy` for the
      # distribution is used.
      packaging_policy=policy,

      # If no argument passed, the default `PythonInterpreterConfig` is used.
      config=python_config,
)
exe.write_aggregate_license_text("licenses.md", "This software contains software subject to licenses as described below.")
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2514,6 +2514,7 @@ dependencies = [
  "tugger-common",
  "tugger-file-manifest",
  "tugger-licensing",
+ "tugger-licensing-net",
  "tugger-rust-toolchain",
  "tugger-windows",
  "tugger-wix",
--- a/pyoxidizer/Cargo.toml
+++ b/pyoxidizer/Cargo.toml
@@ -100,6 +100,11 @@ path = "../tugger-file-manifest"
 version = "0.5.0-pre"
 path = "../tugger-licensing"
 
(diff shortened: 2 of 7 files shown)


pajowu:
Hey, I just wanted to check in again to ask if I can do anything to help evaluating this PR. I completely unterstand if you just didn't have time to review it yet, but maybe there is something where I can help you with this?

indygreg:
Sorry it took so long, but I've applied this commit locally and should push it (along with some heavily refactoring to license handling) sometime in the next few days.

The ability to write a unified licenses file should make it into the next PyOxidizer release.

Thank you for your contribution!

pajowu:
I just noticed that you included the changes and tried the new version. It works great, thank you a lot!

Issue #439, opened 2021-08-30
pyoxidizer not using `src/main.rs` and `build.rs` files gernerated from `init-rust-project` command


hamzamohdzubair:
Below is my `src/main.rs` file. For some reason i am unable to run any `println!()` macro inside my main function in `src/main.rs`.

My python module is called `one.py` and it only has this line `print("Hello World")` and that prints out perfectly when i run `pyoxidizer run` but rust `println!()` macros in `src/main.rs` don't print out.

I have only made the following changes to the standard project generated by `pyoxidizer init-rust-project pytest1`:
1. uncommented relevant lines in `pyoxidizer.bzl`
2. created a file called one.py and added the line: `print("Hello World")`
3. added `println!()` macros to `src/main.rs` file

If any more information is required, i will provide that too.

## src/main.rs
```rust
#![windows_subsystem = "console"]

use pyembed::{MainPythonInterpreter, OxidizedPythonInterpreterConfig};

// Various cargo features can be defined to install a custom global allocator
// for Rust.
//
// Note that this *only* controls Rust's allocator: the Python interpreter
// has its own memory allocator settings on the
// `pyembed::OxidizedPythonInterpreterConfig` that will need to be set in
// order to fully leverage a custom allocator.

#[cfg(feature = "global-allocator-jemalloc")]
#[global_allocator]
static GLOBAL: jemallocator::Jemalloc = jemallocator::Jemalloc;

#[cfg(feature = "global-allocator-mimalloc")]
#[global_allocator]
static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;

#[cfg(feature = "global-allocator-snmalloc")]
#[global_allocator]
static GLOBAL: snmalloc_rs::SnMalloc = snmalloc_rs::SnMalloc;

// Include an auto-generated file defining a
// `fn default_python_config<'a>() -> pyembed::OxidizedPythonInterpreterConfig<'a>`
// which returns an `OxidizedPythonInterpreterConfig` derived by the PyOxidizer
// configuration file.
//
// If you do not want your application to use this generated file or wish
// to explicitly instantiate the `OxidizedPythonInterpreterConfig` used to
// initialize the embedded Python interpreter, simply remove this line and
// the call to `default_python_config()` below.
include!(env!("PYOXIDIZER_DEFAULT_PYTHON_CONFIG_RS"));

fn main() {


    println!("rust::main.rs::main");

[...]

PR #466, opened 2021-10-20
write extra_files when building pyembed artifacts?

dae:
Intended for discussion, not merging as-is. Based on the 0.17 branch; I'm having issues with the main branch which I'll look into more later.

I'd like to be able customize the icon of the produced .exe file on Windows, and run some custom Rust code before the Python interpreter is loaded. I ran into some problems along the way.

I started by using `pyoxidizer init-rust-project sample`. It tells me:

```
A new Rust binary application has been created in sample

This application can be built by doing the following:

  $ cd sample
  $ pyoxidizer build
  $ pyoxidizer run

[...]
```

So I add a print statement to the top of main.rs, and run build/run. Observations:

- run rebuilds everything again. In the main branch the build line appears to have been removed when creating a normal project, but is still shown when using init-rust-project.
- when run, the print statement doesn't appear.

After digging into the sources, I find that pyoxidizer will unconditionally create an ephemeral project when building, ignoring any sources in the current folder. Looks like this was previously reported in #439. Ok, I'll need to try the direct-with-cargo approach instead.

```
% zstdcat /Users/dae/Library/Caches/pyoxidizer/python_distributions/cpython-3.9.6-x86_64-apple-darwin-pgo-20210724T1424.tar.zst | tar xf -
% cd sample && CARGO_TARGET_DIR=target PYOXIDIZER_EXECUTABLE=$HOME/.cargo/bin/pyoxidizer PYTHON_SYS_EXECUTABLE=$HOME/python/install/bin/python3.9 PYOXIDIZER_CONFIG=$(pwd)/pyoxidizer.bzl cargo build --features cpython-link-unresolved-static,allocator-jemalloc
[...]
% ./target/debug/sample
[src/main.rs:37] "test" = "test"
```

Ok, that's promising. Now let's add an extension module into the equation:

```
    policy.resources_location_fallback = "filesystem-relative:prefix"
    python_config.run_command = "import google.protobuf.internal._api_implementation"
    exe.add_python_resources(exe.pip_download(["protobuf"]))
``` 

I rebuild and try again:

```
dae@dip:~/sample% ./target/debug/sample
[src/main.rs:37] "test" = "test"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: dlopen(/Users/dae/sample/target/debug/prefix/google/protobuf/internal/_api_implementation.cpython-39-darwin.so, 2): image not found
zsh: exit 1     ./target/debug/sample
dae@dip:~/sample% find . -name _api_implementation.cpython-39-darwin.so
dae@dip:~/sample%
```

Hmm, it seems extra_files are nowhere to be found. Back to digging through the source code. It looks like to_embedded_resources() ignores extra_files - they only seem to handled when
building an executable.

With the patch in this PR, I am able to successfully change the icon by adding it to the resources file. But I presume there are other cases where the extra_files should not be included. Which leads me to a few questions:

- Is there a better way to modify the default Rust/resources that I've missed?
- If not, would there be any interest in a change like this? Presumably it would need to be
gated by an env var?

### DIFF
--- a/pyoxidizer/src/project_building.rs
+++ b/pyoxidizer/src/project_building.rs
@@ -484,6 +484,10 @@ pub fn build_pyembed_artifacts(
         ))? {
             let p = p?;
 
+            // leave the extra files in the build dir; we do not need them to build the binary
+            if p.file_name() == "extra_files" {
+                continue;
+            }
             let dest_path = artifacts_path.join(p.file_name());
             std::fs::copy(&p.path(), &dest_path).context(format!(
                 "copying {} to {}",
--- a/pyoxidizer/src/py_packaging/binary.rs
+++ b/pyoxidizer/src/py_packaging/binary.rs
@@ -746,6 +746,13 @@ impl<'a> EmbeddedPythonContext<'a> {
         Ok(())
     }
 
+    /// Write extra files like dynamic extension modules to the build folder.
+    pub fn write_extra_files(&self, dest_dir: impl AsRef<Path>) -> Result<()> {
+        self.extra_files
+            .materialize_files(dest_dir.as_ref().join("extra_files"))?;
+        Ok(())
+    }
+
     /// Write out files needed to build a binary against our configuration.
     pub fn write_files(&self, dest_dir: &Path) -> Result<()> {
         self.write_packed_resources(&dest_dir)
@@ -756,6 +763,8 @@ impl<'a> EmbeddedPythonContext<'a> {
             .context("write_interpreter_config_rs()")?;
         self.write_pyo3_config(&dest_dir)
             .context("write_pyo3_config()")?;
+        self.write_extra_files(&dest_dir)
+            .context("write_extra_files()")?;
 
         Ok(())
     }


dae:
Ok, I've rebased this over the main branch, and confirmed it worked on the Mac and Windows machines I tested when used with the commands mentioned on https://github.com/indygreg/PyOxidizer/pull/467/files

I hit a few stumbling blocks when trying to get 0.18 working:

- I first tried with standalone build mode and PYO3_PYTHON set to a standalone Python. It failed with:

```
  = note: ld: warning: directory not found for option '-L/install/lib'
          ld: library not found for -lpython3.9
          clang: error: linker command failed with exit code 1 (use -v to see invocation)


error: could not compile `python-oxidized-importer` due to previous error
```

Presumably pyo3 was trying to autodetect the Python layout, and failing. I was able to work around it by generating the artifacts in advance - is that the intended workflow now?

- I saw mention in the change notes of nightly no longer being required, but it seems that's not the case when building pyoxidizer itself - I ran into an issue that looks like https://github.com/dtolnay/syn/issues/776#issuecomment-863234856 when trying to build with 1.55 on Windows

PR #467, opened 2021-10-20
add an example of how to compile a Rust project

dae:
It took me quite a few hours to figure out how to get a project building with cargo on 0.17, and then figuring out how to get the latest master branch working took time again. Hopefully these small changes make things easier for other people :-)

Related: #466, #439


indygreg (approved):
This seems mostly accurate. There are a few pieces that could be tweaked a little (such as encouraging use of `PYO3_CONFIG_FILE` instead). I'll likely commit this as-is or close to as-is then modify the text slightly in follow-up commits. This will involve me rebasing, so GitHub may not mark this PR as merged due to the Git commit SHA changing.

Thank you for taking the time to contribute your solution to the docs!

Issue #312, opened 2020-11-05
Executable cannot start without PythonPackagingPolicy.include_distribution_sources=True


wkschwartz:
I'm using PyOxidizer v0.9.0 on macOS.

# Reproducing the actual bug

Place the following in `pyoxidizer.bzl` in an otherwise empty directory and execute `pyoxidizer run`.

```bzl
def make():
    dist = default_python_distribution()
    python_config = dist.make_python_interpreter_config()
    python_config.config_profile = "python"
    python_config.run_mode = "eval:import sys; from pprint import pprint; pprint(sys.argv)"

    policy = dist.make_python_packaging_policy()
    policy.set_resource_handling_mode("classify")
    policy.include_distribution_sources = False
    policy.bytecode_optimize_level_zero = True
    policy.include_distribution_resources = True
    exe = dist.to_python_executable(
        name="pyapp",
        packaging_policy=policy,
        config=python_config,
    )
    files = FileManifest()
    files.add_python_resource(".", exe)
    return files
register_target("install", make, default=True)
resolve_targets()
```

It compiles just fine. When it runs, it produces the following error message (paths shortened for clarity; also, despite the output below, my current shell session defines no `PYTHON*` environment variables):

```
Python path configuration:
  PYTHONHOME = './build/x86_64-apple-darwin/debug/install'
  PYTHONPATH = (not set)
  program name = './build/x86_64-apple-darwin/debug/install/./pyapp'
  isolated = 0
  environment = 1
  user site = 1
  import site = 1
  sys._base_executable = './build/x86_64-apple-darwin/debug/install/./pyapp'
  sys.base_prefix = './build/x86_64-apple-darwin/debug/install'
  sys.base_exec_prefix = './build/x86_64-apple-darwin/debug/install'
  sys.executable = './build/x86_64-apple-darwin/debug/install/./pyapp'
  sys.prefix = './build/x86_64-apple-darwin/debug/install'
  sys.exec_prefix = './build/x86_64-apple-darwin/debug/install'
  sys.path = [
    './build/x86_64-apple-darwin/debug/install/lib/python38.zip',
    './build/x86_64-apple-darwin/debug/install/lib/python3.8',
    './build/x86_64-apple-darwin/debug/install/lib/python3.8/lib-dynload',
  ]
during initializing Python main: init_fs_encoding: failed to get the Python codec of the filesystem encoding
```

[...]

PR #481, opened 2021-11-30
pyembed: GIL should be acquired later in the initialization sequence after checking the return value of _Py_InitializeMain()

ntamas:
The current initialization sequence in `MainPythonInterpreter::init()` tries to acquire the GIL before checking the return value of `_Py_InitializeMain()`, leading to an assertion failure if `_Py_InitializeMain()` fails (which can happen when building a project with `include_distribution_sources = False`, as outlined in #312 ). This PR moves the result check before the `acquite_gil()` call, which provides a more meaningful error message instead of an assertion failure:

`error instantiating embedded Python interpreter: during initializing Python main: init_fs_encoding: failed to get the Python codec of the filesystem encoding`

Note that this error message was present in PyOxidizer 0.17 and before; it changed to an assertion failure somewhere in newer versions. This PR simply restores the original error message.

### DIFF
--- a/pyembed/src/interpreter.rs
+++ b/pyembed/src/interpreter.rs
@@ -190,16 +190,15 @@ impl<'interpreter, 'resources> MainPythonInterpreter<'interpreter, 'resources> {
         // importlib. And if the custom importlib bytecode was registered above,
         // our extension module will get imported and initialized.
         let status = unsafe { pyffi::_Py_InitializeMain() };
-
-        let gil = Python::acquire_gil();
-        let py = gil.python();
-
         if unsafe { pyffi::PyStatus_Exception(status) } != 0 {
             return Err(NewInterpreterError::new_from_pystatus(
                 &status,
                 "initializing Python main",
             ));
         }
+
+        let gil = Python::acquire_gil();
+        let py = gil.python();
 
         let sys_module = py
             .import("sys")


mxk:
Can this be merged in? Just ran into the same issue with the main branch.

indygreg:
> Can this be merged in? Just ran into the same issue with the main branch.

I periodically take breaks from projects so I can clear my mind and look at things with a fresher perspective. That's no excuse to not comment on PRs, however. I'll look at this today since it is trivial. But I'm not sure when it will get into a release.

indygreg (approved):
This regressed as part of the port to pyo3 in commit 48c97a63e5fbca46300170e3bbab7debecee1970.

I'll rebase and push shortly.

Thank you for the fix! Sorry it took so long to get merged.

ntamas:
Thanks for merging @indygreg ! While you are here, I was wondering if you could also take a look at my comment [here](https://github.com/indygreg/PyOxidizer/issues/312#issuecomment-983109876) -- I believe that this is a genuine bug in PyOxidizer and I'd be willing to send a PR to fix it if you can confirm that it is indeed a genuine bug and not my misunderstanding.

stappersg:
On Wed, Jan 19, 2022 at 07:29:19AM -0800, Maxim Khitrov wrote:
> Can this be merged in? Just ran into the same issue with the main branch.

Merge it in your own fork and update this GH-issue with your milage ...

PR #485, opened 2021-12-08
support ARM Linux builds

dae:
fixes errors like the following:

```
error[E0308]: mismatched types
  --> /home/dae/.cargo/git/checkouts/pyoxidizer-dcad61535898da20/200fbd2/pyembed/src/conversion.rs:21:50
   |
21 |         let o = pyffi::PyBytes_FromStringAndSize(b.as_ptr() as *const i8, b.len() as isize);
   |                                                  ^^^^^^^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
   |
   = note: expected raw pointer `*const u8`
              found raw pointer `*const i8`

error[E0308]: mismatched types
   --> /home/dae/.cargo/git/checkouts/pyoxidizer-dcad61535898da20/200fbd2/pyembed/src/interpreter_config.rs:494:46
    |
494 |         set_config_string_from_path(&config, &config.home, home, "setting home")?;
    |                                              ^^^^^^^^^^^^ expected `i32`, found `u32`
    |
    = note: expected reference `&*mut i32`
               found reference `&*mut u32`
```

The conditional compilation in interpreter_config.rs is a bit awkward
(aarch64 on macOS uses i32, not u32), but I presumed a dependency on the
libc crate was not desirable for pyembed.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -85,6 +85,9 @@ rustflags = ["-C", "link-args=-Wl,-export-dynamic"]
 [target.x86_64-unknown-linux-gnu]
 rustflags = ["-C", "link-args=-Wl,-export-dynamic"]
 
+[target.aarch64-unknown-linux-gnu]
+rustflags = ["-C", "link-args=-Wl,-export-dynamic"]
+
 [target.aarch64-apple-darwin]
 rustflags = ["-C", "link-args=-rdynamic"]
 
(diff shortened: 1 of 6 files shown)


dae:
One curious thing I noticed while testing this - orjson from a venv doesn't end up packaged correctly, and I end up having to rename it afterwards:

        os.rename(
            output_folder / "lib/orjson/orjson/cpython-39-aarch64-linux-gnu.so",
            output_folder / "lib/orjson/orjson.cpython-39-aarch64-linux-gnu.so",
        )

No other packages I'm using seem to have this problem, and it doesn't seem to occur on an amd64 system.

dae:
Fixed in #509

Issue #490, opened 2021-12-29
multiphase initialisation suport of pyoxidizer


franzhaas:
Hi,

This is a reduced example to the issue https://github.com/indygreg/PyOxidizer/issues/489

Thanks to @rgommers to point me to the right direction

I made a minimal example showing how cython modules can break pyoxidizer single file executables on windows.

https://github.com/franzhaas/pyoxidizer_cython_example

when importing fib, the repl crashes.

versions used are documented in pipenv.lock.

I am new to pyoxidizer, and would apreciate if I could get some guidnace how to tackle this. Is this a regression? Has that been working with older versions? 

I do have access to working pyd files, which have been build out of hand crafted c code and will have a look at the difference between this and the cython c code, but this is likely not to work out....

Thanks for the grate tool!

franzhaas:
I did look into that issue further.

I was able to create a pyd file that triggers this issue, without calling cython. The issue manifests itself after __Pyx_copy_spec_to_module passes its last log message.

The pyd files i got working without issues use sinlge phase initialisation, but the cython generated c code uses 2 phase initialisation. 

Did anyone have success with 2 phase initialisation?

franzhaas:
I did try to install the extension as a sdist instead of a wheel, and achieved exactly the same result.

franzhaas:
I did disable the 489 support when building the generated cython code. This created a working minimal example.

The code is in the minimal example.
https://github.com/franzhaas/pyoxidizer_cython_example


This example loads and works (in a view toy function calls). Therefore i think this  issue si truely due to the PEP489 multi phase loading.

I am willing to continue to work on that, but i a m unsure if the best angle of attack would be to add the feature to disable PEP 489 to cython, or to add 489 support to pyoxidizer. I am worried that to fix this inside of pyoxidizer i need a lot of internal knwoledge of the pyhton interpreter, python extentions and pyoxidizer, that i am lacking. On the other hand, adding a compile time option to cython apears to be trivial.

The goal i am actually trying to achieve is to have llvmlite numba and numpy in a sinlge file executeable.

franzhaas:
I tried to follow this description.:
https://pyoxidizer.readthedocs.io/en/stable/oxidized_importer_freezing_applications.html#oxidized-importer-freezing

to try out if i can reproduce the issue with a standard python installation, and just the oxidized-importer, and i still observed a problem with multiphase initialisation.

franzhaas:
This apears to be a use after delete error.

[...]

PR #493, opened 2022-01-06
- deal with borrowed object ref

franzhaas:
according to https://docs.python.org/3/c-api/module.html:

PyModuleDef_Init returns a borrowed ref, while PyModule_Create returns a new ref.

To my understanding, we need to use the from_borrowed_ptr, as the returned reference should keep the object allive.

I have tried a mockup of this test on my local machine using.:
   
    PyObject *module;
    module =  PyModuleDef_Init(&spam_def);
    Py_INCREF(module);
    return module;

Which loads. So it apears to be the case that pyoxidizer should increment the ref count in this situation, which this PR tries to achieve.

I was unable to compile this on my local machine, hopefully CI can help.

This is an atempt on this issue: https://github.com/indygreg/PyOxidizer/issues/490

### DIFF
--- a/python-oxidized-importer/src/importer.rs
+++ b/python-oxidized-importer/src/importer.rs
@@ -152,7 +152,7 @@ fn load_dynamic_library(
     let init_fn: py_init_fn = unsafe { std::mem::transmute(address) };
 
     // Package context is needed for single-phase init.
-    let py_module_initial_handle = unsafe {
+    let py_module = unsafe {
         let old_context = pyffi::_Py_PackageContext;
         pyffi::_Py_PackageContext = name_cstring.as_ptr();
         let py_module = init_fn();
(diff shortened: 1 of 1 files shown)


franzhaas:
I tried the built wheel localy on win10, and iw as able to load the multiphase initialisation c extension module, which i previously could not.

https://github.com/franzhaas/PyOxidizer/actions/runs/1664281312#artifacts

franzhaas:
i managed to run a test case, and the result is that it didnt work

I am still thinking about that...

franzhaas:
https://github.com/franzhaas/numpy_sinlge_file_executable/

here is an example how an extra incref makes a module loadable...
I tried to replicate this extra incref in the importer.rs but failed so far...

franzhaas:
I managed to load the toy multiphase module succesfully

now i work on undoing the too often made incref...

franzhaas:
https://github.com/franzhaas/numpy_sinlge_file_executable/ 

using the loading_numpy tag is able to load numpy and run some numpy functions in the repl.

I dont think that this is the optimal solution.... The way the newref decref is donne looks wrong to me. Still thinking about it...

franzhaas:
this turned out to be a research effort...
The results are in a seperate pull request....

PR #494, opened 2022-01-11
Enable multiphase initialisation using incref

franzhaas:
I encountered issues when loading multi phase initialisation modules, like numpy.random in single file executables.

This patch fixes these issues for me, however i have the impression that this could be written in a better way. The newref/decref pair feels not very elegant.

### DIFF
--- a/python-oxidized-importer/src/importer.rs
+++ b/python-oxidized-importer/src/importer.rs
@@ -199,7 +199,7 @@ fn load_dynamic_library(
         return if py_module.is_null() {
             Err(PyErr::fetch(py))
         } else {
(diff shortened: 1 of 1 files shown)


franzhaas:
I beleive this is a bodge. I belive by keeping the first py_module variable alife for the right length, this can be solved properly.

But I am new to both rust and python internals. So I cant do this myself.

indygreg:
After looking at the bug report in detail, I'm about to push my own commit (with a lot of inline comments explaining the nuances) that should solve the multi-phase initialization bug. So closing this in favor of that.

Thank you again for providing detail to help track down this very subtle bug!

PR #509, opened 2022-02-17
pyembed: Fix types for linux/aarch64

DanielShaulov:
The c_char and wchar_t types on linux/aarch64 are unsigned, so some
casts to "*const i8" were turned to "*const c_char" and the type
definition for wchar_t to be used from libc.
libc was a transient dependency (from pyo3 at least) and is now direct.
Also fixes default_target to return the correct value on linux/aarch64.

### DIFF
--- a/pyoxidizer/src/projectmgmt.rs
+++ b/pyoxidizer/src/projectmgmt.rs
@@ -41,7 +41,11 @@ use {
 pub fn default_target() -> Result<String> {
     // TODO derive these more intelligently.
     if cfg!(target_os = "linux") {
-        Ok("x86_64-unknown-linux-gnu".to_string())
+        if cfg!(target_arch = "aarch64") {
+            Ok("aarch64-unknown-linux-gnu".to_string())
+        } else {
+            Ok("x86_64-unknown-linux-gnu".to_string())
+        }
     } else if cfg!(target_os = "windows") {
         Ok("x86_64-pc-windows-msvc".to_string())
     } else if cfg!(target_os = "macos") {
--- a/.cargo/config
+++ b/.cargo/config
@@ -85,6 +85,9 @@ rustflags = ["-C", "link-args=-Wl,-export-dynamic"]
 [target.x86_64-unknown-linux-gnu]
 rustflags = ["-C", "link-args=-Wl,-export-dynamic"]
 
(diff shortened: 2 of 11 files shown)


dae:
The use of libc makes this cleaner than #485, but you might want to include the config changes in that PR as well.

DanielShaulov:
@dae - Thanks, added.
@indygreg - Anything we can do to help this get merged?

pyoxidizer/src/projectmgmt.rs
```diff
@@ -41,7 +41,11 @@ use {
 pub fn default_target() -> Result<String> {
     // TODO derive these more intelligently.
     if cfg!(target_os = "linux") {
-        Ok("x86_64-unknown-linux-gnu".to_string())
+        if cfg!(target_arch = "aarch64") {
+            Ok("aarch64-unknown-linux-gnu".to_string())
```
indygreg, line 45: This change is correct. However, since we don't yet have default distributions for `aarch64-unknown-linux-gnu` in PyOxidizer, this will result in (cross-compile) builds breaking on aarch64-unknown-linux-gnu today. This is probably acceptable, as I don't think things work very well on aarch64-unknown-linux-gnu today.

Getting Linux aarch64 working is something I'd like to do for the next release. It was hung up behind some issues with python-build-standalone, which should hopefully be resolved with the next refresh of default Python distributions.

indygreg (approved):
This change looks good. Thank you for helping with Linux aarch64 support!

I'll merge this once CI comes back clean.

indygreg:
I rebased and squashed manually. GitHub isn't smart enough to reflect a merge done outside the web UI / API, unfortunately. But rest assured your changes did get merged.

pyoxidizer/src/projectmgmt.rs
DanielShaulov: While the default doesn't work, compiling the `python-build-standalone` repo for aarch64 works and manually setting the `PythonDistribution(local_path=...)` to the resulting file worked for me on an aarch64 machine (at least as far as opening an interpreter), so I don't think you are too far from everything just working :slightly_smiling_face:. I didn't try cross-compiling, just compiling it on an aarch64 host

PR #515, opened 2022-02-23
Der entitlements

### LINKED ISSUE #513: support including entitlements in der format and bump minimum signature version (issues with ios 15.*)
so was very excited getting it to work on ios 14.8.1. later I decided to update to 15.3.1 after running into trouble with `idevicedebug`. now I get an error: `The code signature version is no longer supported.` I checked that it has nothing to do with the embedded.mobileprovision as the hash hasn't changed on the mac and installing applications built with xcode works. Will look into this more tomorrow, but looks like it has to do with the `apple-codesign` crate.

### LINKED ISSUE #514: after signing llvm-objdump fails to parse macho format
before signing:
```
llvm-objdump --macho --headers main

Sections:
Idx Name             Size     VMA              Type
  0 __text           0000003c 0000000100004000 TEXT
  1 __stubs          0000000c 000000010000403c TEXT
  2 __stub_helper    00000024 0000000100004048 TEXT
  3 __cstring        00000013 000000010000406c DATA
  4 __unwind_info    00001038 0000000100004080 DATA
  5 __got            00000008 0000000100008000 DATA
  6 __la_symbol_ptr  00000008 000000010000c000 DATA
  7 __data           00000008 000000010000c008 DATA
  8 __cfstring       00000020 000000010000c010 DATA
  9 __objc_imageinfo 00000008 000000010000c030 DATA
```
after signing:
```
llvm-objdump --macho --headers main
llvm-objdump: error: 'main': truncated or malformed object (load command 4 filesize field in LC_SEGMENT_64 greater than vmsize field)
```

dvc94ch:
This closes #514 and closes #513

### DIFF
--- a/apple-codesign/src/code_directory.rs
+++ b/apple-codesign/src/code_directory.rs
@@ -540,23 +540,8 @@ impl<'a> CodeDirectoryBlob<'a> {
     pub fn adjust_version(&mut self) -> u32 {
         let old_version = self.version;
 
-        let mut minimum_version = CodeDirectoryVersion::Initial;
+        let mut minimum_version = CodeDirectoryVersion::SupportsExecutableSegment;
 
-        if self.scatter_offset.is_some() {
-            minimum_version = CodeDirectoryVersion::SupportsScatter;
-        }
-        if self.team_name.is_some() {
-            minimum_version = CodeDirectoryVersion::SupportsTeamId;
-        }
-        if self.spare3.is_some() || self.code_limit_64.is_some() {
-            minimum_version = CodeDirectoryVersion::SupportsCodeLimit64;
-        }
-        if self.exec_seg_base.is_some()
-            || self.exec_seg_limit.is_some()
-            || self.exec_seg_flags.is_some()
-        {
-            minimum_version = CodeDirectoryVersion::SupportsExecutableSegment;
-        }
         if self.runtime.is_some() || self.pre_encrypt_offset.is_some() {
             minimum_version = CodeDirectoryVersion::SupportsRuntime;
         }
--- a/apple-codesign/src/macho.rs
+++ b/apple-codesign/src/macho.rs
@@ -1135,6 +1135,85 @@ impl<'a> std::fmt::Display for EntitlementsBlob<'a> {
     }
 }
 
+/// Represents a DER encoded Entitlements blob.
+///
+/// A DER entitlements blob contains a DER encoded plist with a dict.
+#[derive(Debug)]
+pub struct EntitlementsDerBlob<'a> {
+    der: Cow<'a, [u8]>,
+}
+
+impl<'a> Blob<'a> for EntitlementsDerBlob<'a> {
+    fn magic() -> u32 {
+        u32::from(CodeSigningMagic::EntitlementsDer)
+    }
+
+    fn from_blob_bytes(data: &'a [u8]) -> Result<Self, AppleCodesignError> {
+        let data = read_and_validate_blob_header(data, Self::magic(), "entitlements der blob")?;
+        Ok(Self { der: data.into() })
+    }
+
+    fn serialize_payload(&self) -> Result<Vec<u8>, AppleCodesignError> {
+        Ok(self.der.to_vec())
+    }
+}
+
+impl<'a> EntitlementsDerBlob<'a> {
+    /// Construct an instance using any plist as the payload.
+    pub fn from_string(s: &(impl ToString + ?Sized)) -> Result<Self, AppleCodesignError> {
+        use plist::Value;
+        use rasn::ber::enc::{Encoder as DerEncoder, Error as DerError};
+        use rasn::prelude::*;
+        use std::collections::BTreeMap;
+        fn der_encode_value(encoder: &mut DerEncoder, value: &Value) -> Result<(), DerError> {
+            match value {
+                Value::Array(array) => encoder.encode_sequence(Tag::SEQUENCE, |encoder| {
+                    for v in array {
+                        der_encode_value(encoder, v)?;
+                    }
+                    Ok(())
+                }),
+                Value::Dictionary(dict) => {
+                    // make sure it's sorted alphabetically
+                    let map = dict.into_iter().collect::<BTreeMap<_, _>>();
+                    encoder.encode_sequence(Tag::new(Class::Context, 16), |encoder| {
+                        for (k, v) in map {
+                            encoder.encode_sequence(Tag::SEQUENCE, |encoder| {
+                                encoder.encode_utf8_string(Tag::UTF8_STRING, k)?;
+                                der_encode_value(encoder, v)?;
+                                Ok(())
+                            })?;
+                        }
+                        Ok(())
+                    })
+                }
+                Value::Boolean(boolean) => encoder.encode_bool(Tag::BOOL, *boolean),
+                Value::Data(data) => encoder.encode_octet_string(Tag::OCTET_STRING, data),
+                Value::Integer(integer) => {
+                    let integer = Integer::from(integer.as_signed().unwrap());
+                    encoder.encode_integer(Tag::INTEGER, &integer)
+                }
+                Value::String(string) => encoder.encode_utf8_string(Tag::UTF8_STRING, string),
+                _ => {
+                    // non exhaustive enum
+                    unimplemented!()
+                }
+            }
+        }
+        let plist: plist::Value = plist::from_bytes(s.to_string().as_bytes())
+            .map_err(|_err| AppleCodesignError::EntitlementsInvalid)?;
+        let der = rasn::der::encode_scope(|encoder| {
+            encoder.encode_sequence(Tag::new(Class::Application, 16), |encoder| {
+                encoder.encode_integer(Tag::INTEGER, &Integer::from(1))?;
+                der_encode_value(encoder, &plist)?;
+                Ok(())
+            })
+        })
+        .map_err(|_err| AppleCodesignError::EntitlementsInvalid)?;
+        Ok(Self { der: der.into() })
+    }
+}
+
 /// A detached signature.
 #[derive(Debug)]
 pub struct DetachedSignatureBlob<'a> {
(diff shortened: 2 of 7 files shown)


apple-codesign/src/code_directory.rs
```diff
@@ -540,23 +540,8 @@ impl<'a> CodeDirectoryBlob<'a> {
     pub fn adjust_version(&mut self) -> u32 {
         let old_version = self.version;
 
-        let mut minimum_version = CodeDirectoryVersion::Initial;
+        let mut minimum_version = CodeDirectoryVersion::SupportsExecutableSegment;
```
indygreg, line 543: I'm not strongly opposed to this change. But could you please provide some justification for it? Ideally as an inline comment.

The reason the code is the way it is is to provide maximum flexibility and allow writing out all versions of code signatures using this crate. This change as is prevents doing that.

apple-codesign/src/macho_signing.rs
```diff
@@ -143,6 +143,7 @@ fn create_macho_with_signature(
                     Ok(SEG_LINKEDIT) => {
                         let mut segment = *segment;
                         segment.filesize = new_linkedit_segment_size as _;
+                        segment.vmsize = new_linkedit_segment_size as _;
```
indygreg, line 146: This change (and the similar one a few lines later) seems unrelated to this PR. Would you mind splitting it into its own PR?

apple-codesign/src/macho.rs
```diff
@@ -1135,6 +1135,85 @@ impl<'a> std::fmt::Display for EntitlementsBlob<'a> {
...
+    }
+}
+
+impl<'a> EntitlementsDerBlob<'a> {
+    /// Construct an instance using any plist as the payload.
+    pub fn from_string(s: &(impl ToString + ?Sized)) -> Result<Self, AppleCodesignError> {
+        use plist::Value;
+        use rasn::ber::enc::{Encoder as DerEncoder, Error as DerError};
+        use rasn::prelude::*;
+        use std::collections::BTreeMap;
```
indygreg, line 1167: All imports should be at the top of the file.

apple-codesign/src/macho.rs
```diff
@@ -1135,6 +1135,85 @@ impl<'a> std::fmt::Display for EntitlementsBlob<'a> {
...
+                Value::Boolean(boolean) => encoder.encode_bool(Tag::BOOL, *boolean),
+                Value::Data(data) => encoder.encode_octet_string(Tag::OCTET_STRING, data),
+                Value::Integer(integer) => {
+                    let integer = Integer::from(integer.as_signed().unwrap());
+                    encoder.encode_integer(Tag::INTEGER, &integer)
+                }
+                Value::String(string) => encoder.encode_utf8_string(Tag::UTF8_STRING, string),
+                _ => {
+                    // non exhaustive enum
+                    unimplemented!()
```
indygreg, line 1199: We should return an `Err` instead of panicking.

indygreg (commented):
Thank you for putting this together.

I'm going to have to page in the DER entitlements implementation before I can give this an adequate review. That might take a while, especially since I'm currently traveling.

Glancing at this PR, the first thing that sticks out is I'm not super thrilled about taking yet another ASN.1 crate. We already have `yasna` and `bcder` in the dependencies. I'd prefer to use one of those so we don't bloat the dependencies unnecessarily.

apple-codesign/src/code_directory.rs
dvc94ch: so the justification is that ios 15+ requires this as the minimum version. not sure if we should make the minimum version configurable instead. Is there a use case for creating old versions? Do you know how old macosx or ios would have to be to not support this version?

apple-codesign/src/macho_signing.rs
indygreg: Upon investigating this closer, there appears to be some nuance as to how `vmsize` is set. Specifically, it looks like the vmsize is rounded up to 8k boundaries or something along those lines.

I'm going to dig more and likely handle this in my own commit.

indygreg:
I started looking at this today.

I'm curious how you implemented DER serialization for the entitlements plist. I don't see any code for this in Apple's open source repositories. Although I clearly see references to functions like `CESerializeCFDictionary()` and `CEManagedContextFromCFData()`.

Did you base your implementation on some existing source code or did you reverse engineer the DER encoding? Also, do we need full serialization support for all plist types? Isn't the entitlements plist a `Dictionary<String, bool>`? (I'm not opposed to implementing full support for serialization. Although if we're going to do that we might want to contribute it to the `plist` crate.)

apple-codesign/src/code_directory.rs
indygreg: Thinking about this more and looking at changes Apple has made for newer OS versions, I think we need to build out better support for targeting a specific code directory version when signing. This likely entails teaching `SigningSettings` to hold a minimum version and then adding code to set reasonable defaults for fields on newer versions.

This is beyond the scope of this PR and is something I'll tackle myself. Hopefully I push some commits in the next few days.

indygreg: I implemented functionality for automatically adjusting the minimum code signature version in commit 83eb276d29d6233533a7730ebb8e52504058b461. As the commit message states, this is probably not comprehensive. But it should hopefully address iOS 15+ issues.

dvc94ch:
I think it's a bit more complicated than `Dictionary<String, bool>`. This is the simple example I used during reverse engineering, extracted from a macho binary. The structure follows what I extracted from the `embedded.mobileprovision` which needs to match.
```
openssl asn1parse -in Entitlements.der -inform der
    0:d=0  hl=3 l= 200 cons: appl [ 16 ]
    3:d=1  hl=2 l=   1 prim: INTEGER           :01
    6:d=1  hl=3 l= 194 cons: cont [ 16 ]
    9:d=2  hl=2 l=  58 cons: SEQUENCE
   11:d=3  hl=2 l=  22 prim: UTF8STRING        :application-identifier
   35:d=3  hl=2 l=  32 prim: UTF8STRING        :N5P5VW9AJG.com.cloudpeer.todoapp
   69:d=2  hl=2 l=  49 cons: SEQUENCE
   71:d=3  hl=2 l=  35 prim: UTF8STRING        :com.apple.developer.team-identifier
  108:d=3  hl=2 l=  10 prim: UTF8STRING        :N5P5VW9AJG
  120:d=2  hl=2 l=  19 cons: SEQUENCE
  122:d=3  hl=2 l=  14 prim: UTF8STRING        :get-task-allow
  138:d=3  hl=2 l=   1 prim: BOOLEAN           :255
  141:d=2  hl=2 l=  60 cons: SEQUENCE
  143:d=3  hl=2 l=  22 prim: UTF8STRING        :keychain-access-groups
  167:d=3  hl=2 l=  34 cons: SEQUENCE
  169:d=4  hl=2 l=  32 prim: UTF8STRING        :N5P5VW9AJG.com.cloudpeer.todoapp
```
```
cat Entitlements.xml
<dict>
    <key>application-identifier</key>
    <string>N5P5VW9AJG.com.cloudpeer.todoapp</string>
    <key>com.apple.developer.team-identifier</key>
    <string>N5P5VW9AJG</string>
    <key>get-task-allow</key>
    <true/>
    <key>keychain-access-groups</key>
    <array>
        <string>N5P5VW9AJG.com.cloudpeer.todoapp</string>
    </array>
</dict>
```

dvc94ch:
one more thing is still missing. this rule constantly bites me, can we just remove it?

```rust
slf.add_rule2(CodeResourcesRule::new("^[^/]+$")?.nested().weight(10));
```

indygreg:
> one more thing is still missing. this rule constantly bites me, can we just remove it?
> 
> ```rust
> slf.add_rule2(CodeResourcesRule::new("^[^/]+$")?.nested().weight(10));
> ```

Resources rule handling is complex. We should discuss this particular problem in another issue. There's a good chance the default rules/regular expressions in this crate are out-of-sync with reality in iOS land. When I initially implemented this crate, I only really focused on macOS. So issues with iOS are not surprising to me. I would like to better support iOS though!

inolen:
Hopefully this question isn't too out of place, but I'm a bit new to the code signing process and was attempting to use --entitlements-xml-path options to encode my entitlements (namely com.apple.security.get-task-allow and com.apple.security.cs.allow-jit). I can see the entitlements when I run `codesign -d --entitlements :- my.app`, but they still don't actually seem to work (can't attach with lldb, MAP_JIT broken).

Should these still be working even when signing without the new DER format?

indygreg:
I pushed a refactor so the signing settings internally store a parsed `plist::Value`. This will make it slightly simpler to normalize the plist to different output formats (XML, DER) via a more central API.

Tonight, I'm going to try to incorporate your DER serialization code and writing of the new blob section. My intent is to change the author annotation on the commits to reflect you as the author. Since you put in the work of reverse engineering this, I want to be sure you get sufficient credit for your work!

indygreg:
I found myself in `bcder` crate hell when trying to implement the DER encoding. I'm warming up to the idea of using `rasn` or some other crate for the ASN.1. `rasn` does seem to be quite nice with its derive macros. You don't want to know how much time it took me to implement ASN.1 data structures in this repo...

indygreg:
I'm about to push some commits that implement DER entitlements writing.

The critical code here was the DER serialization of plists. I broke that out into its own commit and made you the author of it since you deserve the credit.

I also implemented comprehensive tests of the DER serialization, including verifying that `codesign` produces the same DER as we do. Hopefully this will help detect drift between the DER encoding in the future.

I ultimately decided to adopt the `rasn` crate as it does look quite nice. I'm really fond of its `#[derive]` support. I may find myself replacing `bcder` with it in other crates in this repo in time.

Thank you again for contributing the hard work to figuring out DER encoding!

dvc94ch:
> Thank you again for contributing the hard work to figuring out DER encoding!

Well was just a couple hours, and not that hard. Will try your changes soon. After achieving some success on all platforms, I decided to initially focus on android so I can ship the first usable version of [x](https://github.com/cloudpeers/x), and delay having to pay the apple developer tax. Thanks for landing this feature!

PR #516, opened 2022-02-27
Remove py builtin define

DanielShaulov:
This flag causes errors when building the `grpcio` package from source, because it changes what header needs to be included for some internal GC structs. The comment says it is only needed for windows, so checking for that, but the issue probably happens on windows as well, it is just an easy fix for other platforms.


indygreg (approved):
Interesting behavior! I agree limiting the scope of this change to Windows (where it is actually needed) is best.

PR #530, opened 2022-03-21
[pyembed] Add support for `PYTHONPATH` support in `Python` profile when loading resources from filesystem

ddeville:
When using the regular `Python` profile (that is not isolated) one would
expect paths specified in the `PYTHONPATH` environment variable to be
appended to `sys.path`.

This works as expected when using the `in-memory` resources loader.

However, when loading resources from the filesystem in PyOxidizer, one
has to pass the path to the resources in `module_search_paths` and that
leads us to set these paths in the `PyConfig` itself and thus also
setting `module_search_paths_set` to 1, which leads Python to not add
the usual paths to `sys.path` and thus not expanding `PYTHONPATH` into
it either...

This patch reads `PYTHONPATH` manually and adds it to the
`module_search_paths` in `pyembed` whenever the profile is set to
`Python` *and* `module_search_paths` was set in the configuration.

Note that I'm not familiar enough with the codebase to know whether this
is the best place to do this check but it looked like a reasonable place.

### DIFF
--- a/pyembed/src/interpreter_config.rs
+++ b/pyembed/src/interpreter_config.rs
@@ -496,6 +496,21 @@ pub fn python_interpreter_config_to_py_config(
                 "setting module_search_paths",
             )?;
         }
+
+        // Make sure to append `$PYTHONPATH` if the current profile is not isolated since setting
+        // `module_search_paths_set` to 1 means that Python will not expand and add it to
+        // `sys.path` for us.
+        if let PythonInterpreterProfile::Python = value.profile {
+            if let Ok(path) = std::env::var("PYTHONPATH") {
+                for path in std::env::split_paths(&path) {
+                    append_wide_string_list_from_path(
+                        &mut config.module_search_paths,
+                        &path,
+                        "setting module_search_paths",
+                    )?;
+                }
+            }
+        }
     }
     if let Some(executable) = &value.executable {
         set_config_string_from_path(


indygreg (commented):
(Content deleted because I failed to read your PR description.)

indygreg (commented):
If you saw the previous comment I wrote: ignore it. You clearly did your research and I failed to read the PR summary message detailing the state of affairs.

The conversion function you are editing here seems like the most reasonable place for this logic. I just need to grok CPython's logic around how all the config struct fields behave before I merge this. Notably, I want to understand how https://docs.python.org/3/c-api/init_config.html#c.PyConfig.pythonpath_env is involved and whether we should use that instead.

indygreg:
OK, tracing through config reading at initialization time...

`pythonpath_env` is populated from `PYTHONPATH` if `use_environment && pythonpath_env == NULL`. So `PYTHONPATH` should be read if `use_environment` is set. And `use_environment` is set by default by `PyPreConfig_InitPythonConfig()`.

`_PyConfig_InitPathConfig()` is called later in init. This will call `config_init_pathconfig()` if `!module_search_paths_set` or any of `executable`, `prefix`, or `exec_prefix` are NULL. For PyOxidizer, I want to say that this will be called in most situations.

Here's where things get interesting.

`pathconfig_init()` is always called. This calls into `pathconfig_set_from_config()`, which synchronizes the `PyConfig` to a `_PyPathConfig`, copying and setting `module_search_paths` in the process. Critically, after `pathconfig_set_from_config()` copies `module_search_paths`, we call into `_PyPathConfig_Calculate()`, which is the workhorse function for all the logic.

From there, `calculate_path()` calls into `calculate_module_search_path()`, but only if `module_search_paths == NULL`, which it won't be if it was set on `PyConfig`. And `calculate_module_search_path()` is finally what reads from `pythonpath_env`, which is holding the state of `PYTHONPATH`.

So, yeah, since we're setting `PyConfig.module_search_paths`, this effectively preempts `PYTHONPATH` from working. So we need to "proxy" `PYTHONPATH` through `module_search_paths` if both are defined.

indygreg (changes requested):
This is almost ready to go.

We'll definitely want to add tests for this. But running the tests for `pyembed` is wonky. If you don't want to do it, I'll write the tests as part of merging this.

PR #536, opened 2022-03-31
Fix signing of shallow bundles.

dvc94ch:
Last remaining issue for signing ios bundles!


indygreg (approved):
That was easy!

Unfortunately, we still have some major bugs around the handling of frameworks. But hopefully this change is enough to get basic shallow app bundles (such as iOS app bundles) working!

PR #551, opened 2022-04-21
apple-codesign: Enable loading binary as well as XML Info.plists.

ufx:
I've encountered iOS applications using binary `Info.plist`.  This change enables the bundle signer to work with them.

### DIFF
--- a/apple-bundles/src/directory_bundle.rs
+++ b/apple-bundles/src/directory_bundle.rs
@@ -116,7 +116,7 @@ impl DirectoryBundle {
 
         let info_plist_data = std::fs::read(&info_plist_path)?;
         let cursor = std::io::Cursor::new(info_plist_data);
-        let value = plist::Value::from_reader_xml(cursor).context("parsing Info.plist XML")?;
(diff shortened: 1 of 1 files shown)


indygreg (approved):
Ahh, I didn't realize `from_reader()` automatically sniffed the format and parsed accordingly. Nice find!

Note that we may need additional logic to write out a binary plist in the signed bundle or in the embedded signature. I haven't done much testing with iOS apps so I'm curious what your experience is getting these to pass Apple's checks!

ufx:
Thank you!  With my one test case, it works for iOS 14!  I was able to take an app built with an invalid (for my device) signature, analyze metadata with `extract --data cms-pem` and `print-signature-info` to find a good provisioning profile, entitlement & certificate combo, then resign and repackage the app as an `.ipa`.  That package installed and launched with no trouble.  It did fail signing with compiled storyboards, which I circumvented with `--exclude=*.storyboardc`.

This is a great tool to help us consolidate our application instrumentation & resigning services on non-macOS systems.

Issue #571, opened 2022-06-01
[apple-codesign] ParseInt Error


magarcia:
When I try to notarize a `.pkg`, it throws an XML error saying `Error: Apple flat package error: XML error: ParseIntError: invalid digit found in string`. After digging a bit, I found out that comes from the [PackageInfo definition](https://github.com/indygreg/PyOxidizer/blob/72ecaa846f40d2dcb18464c9a9c6d838654bbb7d/apple-flat-package/src/package_info.rs#L78) because the version is typed as `u32` while it could be actually an `string` ([reference](http://s.sudre.free.fr/Stuff/Ivanhoe/FLAT.html)). I had `2.71.1` in the version field when I got this error.

PR #573, opened 2022-06-01
apple-flat-package: make version a string

magarcia:
[PackageInfo version](https://github.com/indygreg/PyOxidizer/blob/72ecaa846f40d2dcb18464c9a9c6d838654bbb7d/apple-flat-package/src/package_info.rs#L78) is typed as `u32` but it could be actually an `string` ([reference](http://s.sudre.free.fr/Stuff/Ivanhoe/FLAT.html)). For example, a semantic version (like `2.12.2`) is a valid version.

Fixes: #571

### DIFF
--- a/apple-flat-package/src/package_info.rs
+++ b/apple-flat-package/src/package_info.rs
@@ -75,7 +75,7 @@ pub struct PackageInfo {
     ///
     /// This is the version of the package itself, not the version of the application
     /// being installed.
-    pub version: u32,
+    pub version: String,
 
     // End of attributes. Beginning of elements.
     #[serde(default)]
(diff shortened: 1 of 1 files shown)


magarcia:
> Good catch!
> 
> Since you've submitted a few PRs for flat package handling, I should say that this code isn't as battle tested as I would like it to be. It wouldn't surprise me if there were a long tail of random bugs with it. Please file issues / PRs as you come across problems! No need to file an issue before the PR.

Cool, thanks. I just filed the issue first because I didn't know if I would be able to find a solution (I'm not very fluent in rust 😅)

PR #575, opened 2022-06-01
apple-flat-package: fix scripts type

magarcia:
I'm not sure if this is the best approach, but it seems to fix the issue described on indygreg/apple-platform-rs#9. Changing the type of `scripts` from `Vec<Script>` to `Option<Script>` and Script from `enum` to `struct`.

### DIFF
--- a/apple-flat-package/src/package_info.rs
+++ b/apple-flat-package/src/package_info.rs
@@ -107,7 +107,7 @@ pub struct PackageInfo {
 
     /// Scripts to run before and after install.
     #[serde(default)]
-    pub scripts: Vec<Script>,
+    pub scripts: Option<Script>,
 
     #[serde(default)]
     pub strict_identifiers: Vec<BundleRef>,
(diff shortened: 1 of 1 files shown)


indygreg (changes requested):
Hmm. This might be a quirk in the XML serialization handling. I would naively expect the code as previously written to work! Perhaps the issue here is it wanted to see multiple `<scripts>`?

Anyway, we obviously need to change something to make this work.

The change of `scripts` from `Vec<Script>` to `Option<Script>` is reasonable for representing a 0 or 1 `<scripts>`.

And changing `enum Script` to `struct Script` is fine.

But since I believe you can have multiple `<preinstall>` and `<postinstall>`, I think we need to change these fields to `Vec<PreInstall>` and ensure they have the `serde(default)` derivation in place so they are empty by default.

Could you please try that?

indygreg:
I've moved `apple-flat-packages` to https://github.com/indygreg/apple-platform-rs. Please rebase this PR and submit it against the new project if you'd still like it to be considered. (The new repository has the full history of https://github.com/indygreg/PyOxidizer so a rebase should work assuming there aren't other code changes since this PR was submitted.)

PR #576, opened 2022-06-01
apple-codesign: add support for component notarization

magarcia:
I'm not completely sure this is covering all the use cases, but it works for us where we have an `Installer.pkg` that contains:
```
Installer.pkg
├── Bom
├── PackageInfo
├── Payload
└── Scripts
    ├── postinstall
    └── preinstall
```

### DIFF
--- a/apple-codesign/src/notarization.rs
+++ b/apple-codesign/src/notarization.rs
@@ -464,9 +464,17 @@ pub fn write_flat_package_to_app_store_package<F: Read + Seek + Debug>(
         id
     } else if let Some(_component) = pkg.root_component()? {
         warn!("notarizing a component installer");
+        
+        // We need to extract the primary component identifier from the `PackageInfo` XML.
+        let id = match _component.package_info() {
+            Some(_pkginfo) => _pkginfo.identifier.clone(),
+            None => {
+                error!("unable to find package identifier in component package (please report this bug)");
+                return Err(AppleCodesignError::NotarizeFlatPackageParse);
+            },
+        };
 
-        error!("support for notarizing a component installer is not yet implemented");
-        return Err(AppleCodesignError::NotarizeFlatPackageParse);
+        id
     } else {
         error!("do not know how to extract bundle identifier from package installer");
         error!("please report this bug");


apple-codesign/src/notarization.rs
```diff
@@ -464,9 +464,17 @@ pub fn write_flat_package_to_app_store_package<F: Read + Seek + Debug>(
         id
     } else if let Some(_component) = pkg.root_component()? {
         warn!("notarizing a component installer");
+        
+        // We need to extract the primary component identifier from the `PackageInfo` XML.
+        let id = match _component.package_info() {
```
indygreg, line 469: Nit: `_component` should be renamed to `component` a few lines above.

I'll fix this in flight.

indygreg (approved):
This seems reasonable to me. But I'd like to verify that component notarization works. I'm thinking that if I left out support for component notarization and implementing it was this easy, surely I had a better reason to do so! But I could have been focused on getting any flat packages to notarize that I left it half baked and forgot to revisit the missing functionality!

Give me a few hours to verify this. I may just push the commit if it seems to work.

Thank you for this and your other PRs!

indygreg:
I'm wondering if the reason I didn't implement this is because I couldn't figure out a way to verify it worked!

So far every component `.pkg` I've thrown at it has failed notarization with:

```
upload log>   "issues": [
upload log>     {
upload log>       "severity": "error",
upload log>       "code": null,
upload log>       "path": "installer.pkg",
upload log>       "message": "The binary is not signed.",
upload log>       "docUrl": null,
upload log>       "architecture": null
upload log>     }
upload log>   ]
```

It sounds like Apple is looking for a signed entity within the component package. The `.pkg` I'm uploading do have signed entities. So I'm curious what's wrong here.

Do you have an example component package I could test with or instructions on producing one that you are able to get notarized with this patch? I could probably figure this out, but I want to be sure we're seeing eye-to-eye and that we both agree that component notarization *can* work at all.

indygreg:
With the adoption of the new Notary API in #593, the old notarization code that was reading the flat packages has gone away, as we now treat the uploaded assets as mostly a black box. So this change should no longer be needed.

If there are remaining issues with component notarization, please file a new issue or PR.

PR #594, opened 2022-06-17
tugger-binary-analysis: Add glibc and gcc versions for newer distro versions

tsibley:
All of these determined by querying package info from inside corresponding container images using commands like:

    docker run --rm -it fedora:33 yum info --installed glibc
    docker run --rm -it fedora:33 yum info --installed libgcc

    docker run --rm -it registry.access.redhat.com/ubi9/ubi yum info --installed glibc
    docker run --rm -it registry.access.redhat.com/ubi9/ubi yum info --installed libgcc

    docker run --rm -it opensuse/leap:15.2 zypper info glibc
    docker run --rm -it opensuse/leap:15.2 zypper info libgcc_s1

    docker run --rm -it debian:11 apt-cache policy libc6
    docker run --rm -it debian:11 apt-cache policy libgcc-s1

    docker run --rm -it ubuntu:22.04 apt-cache policy libc6
    docker run --rm -it ubuntu:22.04 apt-cache policy libgcc-s1

The output of these commands matched for the existing distro versions here, giving me confidence in their correctness for the newer distro versions, with the notable exception of OpenSUSE 15.0 reporting libgcc_s1 8.2.1 for me instead of 7.3.1 as recorded here.

### DIFF
--- a/tugger-binary-analysis/src/linux_distro_versions.rs
+++ b/tugger-binary-analysis/src/linux_distro_versions.rs
@@ -30,10 +30,17 @@ pub static GLIBC_VERSIONS_BY_DISTRO: Lazy<BTreeMap<&'static str, DistroVersion>>
                 ("30", "2.29"),
                 ("31", "2.30"),
                 ("32", "2.31"),
+                ("33", "2.32"),
+                ("34", "2.33"),
+                ("35", "2.34"),
+                ("36", "2.35"),
             ],
         );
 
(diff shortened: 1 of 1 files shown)


tsibley:
I scanned the logs of all the failing CI checks, and it seems to me that they're all unrelated to this change.

indygreg (approved):
Nice. Thank you for refreshing these lists!

Issue #569, opened 2022-05-31
For python 3.10, the name of lib/3.10/site-packages is being truncated to lib/3.1/site-packages


sparkstartconsulting:
The 0 is being dropped from 3.10 when the folder name is being created

(.venv) ➜  rust_call_python git:(master) ✗ pyoxidizer run        
resolving 1 targets
resolving target install
resolving target exe
resolving Python distribution https://github.com/indygreg/python-build-standalone/releases/download/20220227/cpython-3.10.2%2B20220227-x86_64-unknown-linux-gnu-pgo-full.tar.zst (sha256=28ded83c1742870fe0633619ec43f9b3a693078bf7da1e2afb81a543517e032d)
downloading https://github.com/indygreg/python-build-standalone/releases/download/20220227/cpython-3.10.2%2B20220227-x86_64-unknown-linux-gnu-pgo-full.tar.zst
Python distribution available at /home/bmiller3230/.cache/pyoxidizer/python_distributions/cpython-3.10.2%2B20220227-x86_64-unknown-linux-gnu-pgo-full.tar.zst
reading data from Python distribution...
thread 'main' panicked at 'unable to get directory entry: Error { depth: 0, inner: Io { path: Some("./.venv/lib/**python3.1**/site-packages"), err: Os { code: 2, kind: NotFound, message: "No such file or directory" } } }', /home/bmiller3230/.cargo/registry/src/github.com-1ecc6299db9ec823/python-packaging-0.12.0/src/filesystem_scanning.rs:117:35
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

The text of pyoxidizer.bzl file is pasted below:

```
# This file defines how PyOxidizer application building and packaging is
# performed. See PyOxidizer's documentation at
# https://pyoxidizer.readthedocs.io/en/stable/ for details of this
# configuration file format.

# Configuration files consist of functions which define build "targets."
# This function creates a Python executable and installs it in a destination
# directory.
def make_exe():
    # Obtain the default ,/Distribution for our build target. We link
    # this distribution into our produced executable and extract the Python
    # standard library from it.
    dist = default_python_distribution(python_version = "3.10")

    # This function creates a `PythonPackagingPolicy` instance, which
    # influences how executables are built and how resources are added to
    # the executable. You can customize the default behavior by assigning
    # to attributes and calling functions.
    policy = dist.make_python_packaging_policy()

    # Enable support for non-classified "file" resources to be added to
    # resource collections.
[...]

PR #600, opened 2022-07-03
569 fix python version slicing

codifryed:
Hi, this is a fix for #569 

The issue was that the python version was being sliced to exactly 3 characters long, which doesn't work for "3.10". ~~This fix removes the slicing and replaces the leftover parameter use of `version` with `python_major_minor_version()`, as other places already use it like this.~~

~~This fix should work, but does not improve the resiliency of the `resolve_python_paths` function, which could then be incorrectly used in the future. There are of course a few options available to improve that, like copying the `python_major_minor_version` method logic into the mentioned function, or consolidating the logic for both methods somewhere in another function.~~

Since I'm new to the project I'd love to hear your thoughts before going in too deep for a small bug fix. :wink: 

Thanks for this great project btw! :+1: 

Edit: I've gone ahead and consolidated the logic into a new function.

### DIFF
--- a/pyoxidizer/src/py_packaging/standalone_distribution.rs
+++ b/pyoxidizer/src/py_packaging/standalone_distribution.rs
@@ -256,6 +256,16 @@ fn parse_python_json_from_distribution(dist_dir: &Path) -> Result<PythonJsonMain
     parse_python_json(&python_json_path)
 }
 
+fn parse_python_major_minor_version(version: &str) -> String {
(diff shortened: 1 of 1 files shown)


indygreg (approved):
Thank you for the contribution. This change looks good. I'm going to tweak the commit a little locally before pushing (mainly to add info to the release notes).

codifryed:
sounds good :+1:

PR #601, opened 2022-07-04
debian-packaging: Derive hash for PackageVersion

sjoerdsimons:
Useful when e.g. when wanting to stick package name + version in a hash
table.

Signed-off-by: Sjoerd Simons <sjoerd@collabora.com>


indygreg (approved):
Makes sense. Thanks for the contribution!

PR #607, opened 2022-07-12
apple-codesign/README: Fix rcodesign remote install command line

mrmacete:
The previous command line produced the error:

```
error: could not find `rcodesign` in https://github.com/indygreg/PyOxidizer?branch=main with version `*`
```


indygreg (approved):
Thanks for this fix!

PR #614, opened 2022-07-26
Correct sorting and allow signed data without content

ralpha:
- Correct sorting of `signed_attributes`.
- Allow Signed data without content.
- Re-export `Bytes` and `Oid` so `SignedDataBuilder::content_type()` can be used without importing `bytes` and `bcder`.

( The `signed_attributes` Content type is move so they are all grouped together, this makes it easier to read. )

### DIFF
--- a/cryptographic-message-syntax/src/asn1/rfc5652.rs
+++ b/cryptographic-message-syntax/src/asn1/rfc5652.rs
@@ -652,10 +652,18 @@ impl SignedAttributes {
     /// Obtain an instance where the attributes are sorted according to DER
     /// rules. See the comment in [SignerInfo::signed_attributes_digested_content].
     pub fn as_sorted(&self) -> Result<Self, std::io::Error> {
-        // The official rules say you sort by the first element in the container.
-        // That's an Oid, which implements comparisons. So our job is easy.
         let mut res = self.0.clone();
-        res.sort_by(|a, b| a.typ.as_ref().cmp(b.typ.as_ref()));
+        // Ordering on the encoded bytes.
+        // Following: https://stackoverflow.com/a/72492744/2037998
+        res.sort_by_cached_key(|a| {
+            let mut a_encoded = Vec::new();
+            let a_values = a.clone().encode();
+            a_values.write_encoded(Mode::Der, &mut a_encoded).unwrap();
+            a_encoded
+        });
+        // PREVIOUS METHOD: The official rules say you sort by the first element in the container.
+        // That's an Oid, which implements comparisons. So our job is easy.
+        // res.sort_by(|a, b| a.typ.as_ref().cmp(b.typ.as_ref()));
 
         Ok(Self(res))
     }
--- a/cryptographic-message-syntax/Cargo.toml
+++ b/cryptographic-message-syntax/Cargo.toml
@@ -20,6 +20,11 @@ reqwest = { version = "0.11", default-features = false, features = ["blocking", 
 ring = "0.16"
 signature = "1.3"
 
+[features]
+default = []
+# Do not insert content (that was hashed) into `SignedData`
+no_content = []
+
 [dependencies.x509-certificate]
 path = "../x509-certificate"
 version = "0.14.0-pre"
--- a/cryptographic-message-syntax/Cargo.toml
+++ b/cryptographic-message-syntax/Cargo.toml
@@ -20,11 +20,6 @@ reqwest = { version = "0.11", default-features = false, features = ["blocking", 
 ring = "0.16"
 signature = "1.3"
 
-[features]
-default = []
-# Do not insert content (that was hashed) into `SignedData`
-no_content = []
-
 [dependencies.x509-certificate]
 path = "../x509-certificate"
 version = "0.14.0-pre"
--- a/cryptographic-message-syntax/src/lib.rs
+++ b/cryptographic-message-syntax/src/lib.rs
@@ -78,6 +78,8 @@ mod time_stamp_protocol;
 pub use {
     signing::{SignedDataBuilder, SignerBuilder},
     time_stamp_protocol::{time_stamp_message_http, time_stamp_request_http, TimeStampError},
+    bytes::Bytes,
+    bcder::Oid,
 };
 
 use {
@@ -88,7 +90,7 @@ use {
             OID_SIGNING_TIME,
         },
     },
-    bcder::{Integer, OctetString, Oid},
+    bcder::{Integer, OctetString},
     pem::PemError,
     ring::{digest::Digest, signature::UnparsedPublicKey},
     std::{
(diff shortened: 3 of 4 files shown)


cryptographic-message-syntax/src/asn1/rfc5652.rs
```diff
@@ -652,10 +652,18 @@ impl SignedAttributes {
...
-        res.sort_by(|a, b| a.typ.as_ref().cmp(b.typ.as_ref()));
+        // Ordering on the encoded bytes.
+        // Following: https://stackoverflow.com/a/72492744/2037998
+        res.sort_by_cached_key(|a| {
+            let mut a_encoded = Vec::new();
+            let a_values = a.clone().encode();
+            a_values.write_encoded(Mode::Der, &mut a_encoded).unwrap();
+            a_encoded
+        });
+        // PREVIOUS METHOD: The official rules say you sort by the first element in the container.
```
indygreg, line 664: I'm really curious why I left that `first element in the container` comment, which is apparently wrong per your SO discussion. I must have read that somewhere otherwise I wouldn't leave the comment!

indygreg (commented):
Thanks for the PR.

My first reaction is that sorting bug must have been painful to triage. I'm sorry to have to put you through that :/

I wrote this library to implement code signing on Apple platforms and Apple's implementation doesn't seem to emit sorted signed attributes nor does it validate the attributes are sorted when doing signature verification. So that's probably why this bug is present.

As for the code changes, the sorting change looks good. As does the re-exporting of symbols and the minor refactor.

But the introduction of a crate feature will need some work. For that, we'll want a function argument to control behavior. I don't have CMS implementation details paged into my brain at the moment and am not sure what you are trying to do, but there are some APIs (at least on the verification said) for utilizing external/non-inline signed content/data. I want to say the problem here is we don't really provide that API for signature writing.

I think what we may want to do is introduce an enum for signed content:

```rust
enum SignedContent {
   /// No content is being signed.
   None,

   /// Signed content will be embedded in the signature.
   Inline(Vec<u8>),

   /// Signed content is external. Digest is captured in signature.
   External(Vec<u8>),
}
```

Then we add APIs to `SignedDataBuilder` to set `self.signed_content` to the appropriate variant, with the default being the `None` variant. From there, we can conditionally add the digest and inline content when building the DER.

Would that work for your use case?

cryptographic-message-syntax/src/asn1/rfc5652.rs
ralpha: Yeah I also though this was correct. But apparently not. Also changing this actually fixed some issue with the PDF. So Acrobat can not handle the incorrect sorting.
I did look up if I could find more info somewhere about this. But info was very hard to find and/or vague in its description.
Maybe there are different way of doing this because the standard is not so explicit.

This is what I found around this:
> 11.6 Set-of components
> The encodings of the component values of a set-of value shall appear in ascending order, the encodings being compared as
> octet strings with the shorter components being padded at their trailing end with 0-octets.

Source: https://www.itu.int/rec/T-REC-X.690-202102-I/en

But also:
> 9.3 Set components
> The encodings of the component values of a set value shall appear in an order determined by their tags as specified in 8.6 of
> Rec. ITU-T X.680 | ISO/IEC 8824-1. Additionally, for the purposes of determining the order in which components are
> encoded when one or more component is an untagged choice type, each untagged choice type is ordered as though it has a tag
> equal to that of the smallest tag in that choice type or any untagged choice types nested within.

Source: https://www.itu.int/rec/T-REC-X.690-202102-I/en

Although it looks like `SET` is used for some reason `SET OF` sorting is needed.
```
> openssl asn1parse -in signed_content.der -inform DER
    0:d=0  hl=2 l= 105 cons: SET               
    2:d=1  hl=2 l=  24 cons: SEQUENCE          
    4:d=2  hl=2 l=   9 prim: OBJECT            :contentType
   15:d=2  hl=2 l=  11 cons: SET               
   17:d=3  hl=2 l=   9 prim: OBJECT            :pkcs7-data
   28:d=1  hl=2 l=  28 cons: SEQUENCE          
   30:d=2  hl=2 l=   9 prim: OBJECT            :signingTime
   41:d=2  hl=2 l=  15 cons: SET               
   43:d=3  hl=2 l=  13 prim: UTCTIME           :220726144320Z
   58:d=1  hl=2 l=  47 cons: SEQUENCE          
   60:d=2  hl=2 l=   9 prim: OBJECT            :messageDigest
   71:d=2  hl=2 l=  34 cons: SET               
   73:d=3  hl=2 l=  32 prim: OCTET STRING      [HEX DUMP]
```

This looks to be the clearest source I could find.
> SET OF encoding
> A SET OF items is encoded the same way as a SET, including the tag byte of 0x31. For DER encoding, there is a similar requirement that the SET OF must be encoded in ascending order. Because all elements in the SET OF have the same type, ordering by tag is not sufficient. So the elements of a SET OF are sorted by their encoded values, with shorter values treated as if they were padded to the right with zeroes.

https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/#set-of-encoding

I'm by no means an expert on ASN.1 or DER encoding. But it looks like it works and other people (on StackOverflow) say it should be that way. So I guess that is what is needed.
It could also just be a bug in Acrobat (what would not surprise me, there code seems to hanging by a thread.)

Maybe there is a way to switch between the 2 sorting methods using a variable (that we can implement).
But I don't know what would be best in this case.

ralpha:
> My first reaction is that sorting bug must have been painful to triage. I'm sorry to have to put you through that :/

No problem, It was difficult, but was far from the only problem. The PDF spec and all related/used specs are difficult in general.
Having this library, already helped a lot, creating it would be MUCH more difficult.

I was already thinking that the feature was not the best way to do it. But I fixed this now and it looks much better.
Let me know if there is anything else.

cryptographic-message-syntax/src/asn1/rfc5652.rs
indygreg: Thanks for the research! But I'm having a hard time believing I translated any of that into my comment of `sort by the first element in the container`! That being said, I didn't have a great understanding about ASN.1 when I started implementing this crate. So it's quite possible I did something very foolish like interpret `order determined by their tags` to somehow mean `tag = OID`. 🤷‍♂️ 

It doesn't really matter. What does matter is the current sorting is objectively wrong. So we should fix it.

indygreg (approved):
This code looks pretty good!

There are a few things I'd like to change before this merges. Instead of going through indefinite rounds of PR revisions, I'm going to cherry-pick and amend your commits locally. I'll still retain your authorship info in the commits, as you deserve credit for the work. Look for me to push these changes today.

Thanks again for contributing the fix and the new feature to better support signature generation!

PR #625, opened 2022-08-13
Update `libtest-mimic` to 0.5.0

LukasKalbertodt:
I just released [a new version of libtest-mimic](https://github.com/LukasKalbertodt/libtest-mimic/releases/tag/v0.5.0). This is one of the bigger projects using it, so as a sanity check whether the new version still works for real world users, I just did this update. It allowed to remove some of the code here as the API changed a bit.

Two things to note:
- `libtest-mimic` requires Rust 1.58. 
- I do not know anything about this project at all and did not test anything except compiling and running the tests (via `cargo run -p pyoxy`). Please review carefully and/or test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3047,14 +3047,13 @@ dependencies = [
 
 [[package]]
 name = "libtest-mimic"
-version = "0.4.1"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc195aab5b803465bf614a5a4765741abce6c8d64e7d8ca57acd2923661fba9f"
+checksum = "79529479c298f5af41375b0c1a77ef670d450b4c9cd7949d2b43af08121b20ec"
 dependencies = [
  "clap 3.2.16",
- "crossbeam-channel",
- "rayon",
  "termcolor",
+ "threadpool",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


indygreg (approved):
This looks good assuming CI comes back clean. Our current MSRV is 1.60, so 1.58 is no problem.

Thanks for `libtest-mimic` and for the PR!

FWIW the reason I use `libtest-mimic` is so I can use `nextest-rs` locally. But according to https://github.com/nextest-rs/datatest-stable/pull/5 it sounds like we need to wait for some releases before this 0.5 release works with nextest. So I'm going to hold off merging this PR so I don't break my workflow...

LukasKalbertodt:
> assuming CI comes back clean.

It did not, but it doesn't look like this PR is responsible for the failures (e.g. [this](https://github.com/indygreg/PyOxidizer/runs/7831628240?check_suite_focus=true) and [this](https://github.com/indygreg/PyOxidizer/runs/7831627701?check_suite_focus=true)).

Regarding nextest: yep, 0.5.0 doesn't work with nextest due to an oversight on my part. But 0.5.2 should work completely; @sunshowers tested it [here](https://github.com/LukasKalbertodt/libtest-mimic/pull/18). I updated the commit in this PR to require 0.5.2.

PR #628, opened 2022-08-23
Parse info.plist before importing macho settings.

### LINKED ISSUE #627: [apple-codesign] Signing a fat binary may result in different binary identifier in csreq
Hello,

When I sign an application bundle containing a single fat binary (x86_64 and arm64) with rcodesign, I end up with a fat binary whose slices have different binary identifier in their csreq:

- The x86_64 slice (which comes first) has the correct bundle identifier, derived from the Info.plist
- The aarch64 slice (which comes second) has a weird identifier that looks like `<filename>-<hash>`, which I suspect comes from the ad-hoc signature generated by the linker.

Here's a quick reproducer:

```
echo "int main() { return 0; }" > test.c
clang -target x86_64-apple-darwin test.c -o test-x64
clang -target arm64-apple-darwin test.c -o test-arm64
lipo -create -output test test-x64 test-arm64

mkdir -p Test.app/Contents/MacOS
echo -n 'APPL????' > Test.app/Contents/PkgInfo
cp test Test.app/Contents/MacOS/test

cat > Test.app/Contents/Info.plist  << END
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>CFBundleExecutable</key>
	<string>test</string>
	<key>CFBundleIdentifier</key>
	<string>quick.test</string>
	<key>CFBundleInfoDictionaryVersion</key>
	<string>6.0</string>
	<key>CFBundleName</key>
	<string>test</string>
	<key>CFBundlePackageType</key>
	<string>APPL</string>
</dict>
</plist>
END

rcodesign sign Test.app
rcodesign print-signature-info Test.app | grep identifier
```

The final command has this output:

```
          identifier: quick.test
          identifier: test-arm64
```

Whereas I expect the following:

```
          identifier: quick.test
          identifier: quick.test
```

Given that's the identifier specified in the Info.plist.

roblabla:
The Info.plist binary identifier has higher precedence than the one from the MachO binary. For this to work, however, it needs to be set in the settings *before* calling import_settings_from_macho. Fixes indygreg/apple-platform-rs#12

### DIFF
--- a/apple-codesign/src/bundle_signing.rs
+++ b/apple-codesign/src/bundle_signing.rs
@@ -578,8 +578,6 @@ impl SingleBundleSigner {
 
             let mut settings = settings.clone();
 
-            settings.import_settings_from_macho(&macho_data)?;
-
             // The identifier for the main executable is defined in the bundle's Info.plist.
             if let Some(ident) = self
                 .bundle
@@ -591,6 +589,8 @@ impl SingleBundleSigner {
             } else {
                 info!("unable to determine binary identifier from bundle's Info.plist (CFBundleIdentifier not set?)");
             }
+
+            settings.import_settings_from_macho(&macho_data)?;
 
             settings.set_code_resources_data(SettingsScope::Main, resources_data);
 


indygreg:
I've moved `apple-codesign` to https://github.com/indygreg/apple-platform-rs. Please rebase this PR and submit it against the new project if you'd still like it to be considered. (The new repository has the full history of https://github.com/indygreg/PyOxidizer so a rebase should work assuming there aren't other code changes since this PR was submitted.)

PR #640, opened 2022-09-14
apple-bundles: fix iOS framework signing

kabiroberai:
iOS frameworks contain the Info.plist file in their root directory, rather than inside Resources/.


indygreg:
I've moved `apple-bundles` to https://github.com/indygreg/apple-platform-rs. Please rebase this PR and submit it against the new project if you'd still like it to be considered. (The new repository has the full history of https://github.com/indygreg/PyOxidizer so a rebase should work assuming there aren't other code changes since this PR was submitted.)

PR #700, opened 2023-04-24
Make PyOxidizer build local rust project when it can find toml file

JulianZhang:
I am planning to mix  Rust and Python in a new project. When I tried to build the demo project with `cargo build` , I found that the `add_python_resource` step was missing, causing some dependency files to be missing when running Python scripts. Using `pyoxidizer build` can create the correct Python runtime environment, but it always compiles the demo Rust project. 

I found that the easiest solution is to modify some of the PyOxidizer code so that when there is a local project, it compiles the local project instead of creating a new demo. The current condition I have chosen is to find a `toml` file in the current working directory (CWD).

Can this change be considered for merging into the main branch?

### DIFF
--- a/pyoxidizer/src/environment.rs
+++ b/pyoxidizer/src/environment.rs
@@ -259,7 +259,7 @@ pub struct Environment {
     ///
     /// Should only be set if we're running a binary from a `cargo` command,
     /// as the environment variables this keys off of are set by `cargo`.
-    pub cargo_target_directory: Option<PathBuf>,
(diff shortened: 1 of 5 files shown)


JulianZhang:
@indygreg  Hi, I thank PyOxidizer is ready to build  python env with  user's rust project. And I also find there is another PR with same requirement [468](https://github.com/indygreg/PyOxidizer/pull/468).

But the 468 was blocked because failed in checks. If you thank the solution by   468 (use env) is better, I can write a new PR  use that way, and fix all checking fails.

PR #717, opened 2023-09-19
Allow persistant pyemed runtime projects

### LINKED ISSUE #335: pyoxidizer init-rust-project does not works as documented
`
julien@x270:~/test/oxidizer$ pyoxidizer init-rust-project pyapp
     Created binary (application) package
writing pyapp/.cargo/config
writing pyapp/build.rs
writing pyapp/src/main.rs
writing pyapp/pyoxidizer.bzl
writing pyapp/pyapp.exe.manifest
writing pyapp/pyapp-manifest.rc

A new Rust binary application has been created in pyapp

This application can be built by doing the following:

  $ cd pyapp
  $ pyoxidizer build
  $ pyoxidizer run

The default configuration is to invoke a Python REPL. You can
edit the various pyoxidizer.*.bzl config files or the main.rs 
file to change behavior. The application will need to be rebuilt 
for configuration changes to take effect.

julien@x270:~/test/oxidizer$ cd pyapp

`
I modified src/main.rs :

`
julien@x270:~/test/oxidizer/pyapp$ cat src/main.rs 
#![windows_subsystem = "console"]

use pyembed::MainPythonInterpreter;

// Include an auto-generated file containing the default
// `pyembed::OxidizedPythonInterpreterConfig` derived by the PyOxidizer
// configuration file.
//
// If you do not want to use PyOxidizer to generate this file, simply
// remove this line and instantiate your own instance of
// `pyembed::OxidizedPythonInterpreterConfig`.
include!(env!("PYOXIDIZER_DEFAULT_PYTHON_CONFIG_RS"));

fn main() {
    // The following code is in a block so the MainPythonInterpreter is destroyed in an
    // orderly manner, before process exit.
    let code = {
        // Load the default Python configuration as derived by the PyOxidizer config
        // file used at build time.
        let config = default_python_config();

        // Construct a new Python interpreter using that config, handling any errors
        // from construction.
        match MainPythonInterpreter::new(config) {
            Ok(mut interp) => {
                // And run it using the default run configuration as specified by the
                // configuration. If an uncaught Python exception is raised, handle it.
                // This includes the special SystemExit, which is a request to terminate the
                // process.
                // interp.run_as_main()
                let py = interp.acquire_gil().unwrap();
                match py.eval( "print('hello, world')", None, None) {
                    Ok(_) => println!("python code executed successfully"),
                    Err(e) => println!("python error: {:?}", e),
                }
                0
            }
            Err(msg) => {
                eprintln!("{}", msg);
                1
            }
        }
    };

    // And exit the process according to code execution results.
    std::process::exit(code);
}
`

but when running 

`pyoxidizer run`

I get the same python REPL as without modification !

Skehmatics:
Allow the use of pyoxidizer's build system in a project created via `init-rust-project`.

This adds a `runtime` option to `PythonDistribution.to_python_executable()` that specifies the path of the Rust project, and defines it during `init-rust-project`

With this, applications that are happy to continue utilizing pyoxidizer's python resource discovery system can continue doing so while still allowing custom Rust code to be included in the final executable.

Closes #335

### DIFF
--- a/pyoxidizer/docs/pyoxidizer_config_type_python_distribution.rst
+++ b/pyoxidizer/docs/pyoxidizer_config_type_python_distribution.rst
@@ -84,7 +84,7 @@
         The policy automatically uses settings globally appropriate for the
         distribution.
 
-    .. py:method:: to_python_executable(name: str, packaging_policy: PythonPackagingPolicy, config: PythonInterpreterConfig) -> PythonExecutable
+    .. py:method:: to_python_executable(name: str, packaging_policy: PythonPackagingPolicy, config: PythonInterpreterConfig, runtime: str) -> PythonExecutable
 
         This method constructs a :py:class:`PythonExecutable` instance. It
         essentially says *build an executable embedding Python from this
@@ -106,6 +106,12 @@
            The default configuration of the embedded Python interpreter.
 
            Default is what :py:meth:`make_python_interpreter_config` returns.
+
+        ``runtime``
+           The path to a pyoxidizer Rust project that will invoke pyembed.
+           See :ref:`rust_projects` for more on the expected composition of Rust projects.
+
+           If not supplied, a temporary Rust project will be created.
 
         .. important::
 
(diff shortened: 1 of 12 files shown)


chris-laplante:
Very timely, thank you. I just started looking at PyOxidizer this week and thought I was going crazy when I couldn't get the persistent build to work.

chris-laplante:
Commenting back to say that this patch worked fine for me. Thanks! @indygreg would love to see this get merged.

PR #739, opened 2024-03-08
Update python distribution script, add it to justfile and update default distributions

SRv6d:
This PR updates the default python distributions, adding python 3.11 and 3.12, while making the update script easier to use with `just`.


jokroese:
Is there anything we can do to get this over the line? 💪🏼

ntamas:
A lot more work had to be done on PyOxidizer's side to make newer Python distributions work and to ensure that all the tests pass. I've done most of the work in my fork here: https://github.com/ntamas/PyOxidizer/tree/update-python-distributions

